PR #10, opened 2019-01-28
dom-api-task done

Roka20012:
DOM

[Demo](https://roka20012.github.io/dom_practical_task/) | [Code base](https://github.com/Roka20012/kottans_frontend/tree/master/practical_task/practic_dom)

Please, review.


Oodmincheg:
Nice job! Please check your code for resolution 2560x1080. Site doesn't  look like it should be on my monitor :) You can emulate any resolution in yours browser developer tools. Don't hesitate ask something here if you need.

Roka20012:
@Oodmincheg Thanks :) I fix it

submissions/Roka20012/js-dom/interactive.js
```diff
@@ -2,52 +2,49 @@
...
+                <span>Rem ipsum eligendi id consequatur, expedita cumque neque. Dolor natus tempora repellendus quaerat accusamus eligendi doloribus, accusantium dolore voluptatibus dolorem fugiat architecto dolorum inventore error! Magnam doloremque velit, commodi minima!</span></p>
+                <p><span>Eligendi architecto voluptas alias labore reprehenderit temporibus sed, et laborum repellendus quis quae dicta mollitia voluptates, hic? Excepturi praesentium vel, labore veniam earum facere quae deserunt eius quaerat deleniti quod.</span>
+                <span>Qui asperiores assumenda, magni hic magnam quis earum blanditiis dolor adipisci, quibusdam voluptatibus ducimus distinctio quidem eaque dolorum voluptate delectus fuga nisi beatae veritatis. Debitis optio aspernatur deserunt commodi! Saepe.</span></p>`,
     4: `<p><span><strong>Storie 4.</strong> Lorem ipsum dolor sit amet, consectetur adipisicing elit. Odit similique molestiae eligendi laudantium minus consequatur repellat ipsa, nemo repudiandae perspiciatis, magni aut temporibus nam ipsum expedita aperiam laboriosam quia, sapiente.</span>
```
Oodmincheg: Can you try to remove `<strong>` element to your css?

submissions/Roka20012/js-dom/interactive.js
```diff
@@ -2,52 +2,49 @@
...
+                <span>Hic nesciunt nemo quia odio quo numquam cum praesentium quis dolor impedit amet porro aut eos eaque dolore esse incidunt quos, voluptas doloremque eveniet blanditiis! Earum doloremque temporibus cupiditate laboriosam.</span>
+                <span>Doloribus fugiat cumque nostrum, error iusto vitae voluptatem porro ipsam officiis provident ut ad. Repellat quas at temporibus numquam rem culpa adipisci, blanditiis soluta vitae accusantium, veritatis laboriosam. Voluptate, praesentium.</span>
+                <span>Laborum rid, quasi voluptate animi nulla, molestias provident possimus, architecto accusamus magnam veniam rerum perspiciatis dicta assumenda unde et eveniet sed illum quaerat aut eum suscipit! Deserunt voluptas, doloribus.</span>
+                <span>Rem ipsum eligendi id consequatur, expedita cumque neque. Dolor natus tempora repellendus quaerat accusamus eligendi doloribus, accusantium dolore voluptatibus dolorem fugiat architecto dolorum inventore error! Magnam doloremque velit, commodi minima!</span></p>
```
Oodmincheg: Are you really need so many `<span>` in your markup? If yes, can you explain me why?

submissions/Roka20012/js-dom/interactive.js
```diff
@@ -2,52 +2,49 @@
...
-        document.documentElement.clientWidth ||
-        document.body.clientWidth;
-    let maxWidth = 900;
-    if (width < maxWidth) {
+function hideBurgerMenu(target) {
+    let computedStyle = getComputedStyle(burger);
+
+    if (computedStyle.left == "15px") {
         nav.classList.toggle("menu-open");
         nav.classList.toggle("menu-close");
```
Oodmincheg: Maybe you can use one class here(only 'open' for example)?

submissions/Roka20012/js-dom/interactive.js
```diff
@@ -92,14 +89,14 @@ function checkStory({ target }) {
     if (target.tagName == "A") {
         for (let i = 0; i < nav.children.length; i++) {
```
Oodmincheg: It isn't a good idea use directly tagName in you JS. Imagine if you would need to add another `<a>` element to your html. and then another. all of them inerhit this logic. Add class please.

submissions/Roka20012/js-dom/interactive.js
```diff
@@ -0,0 +1,118 @@
...
+        nav.classList.toggle("menu-open");
+        nav.classList.toggle("menu-close");
+        burger.classList.toggle("open");
+    }
+}
+
+/**
+ * додає колір для вибраного елемента
+ * author: Rostyk Broslavsky
+ * email: broslavsky.rostislaw@gmail.com
```
Oodmincheg: It isn't a good practice left you comments in language different than English(on github repos especially). For this task you can left like it is. But in feature try to use English. Actually you can try write code without comments using good names of functions in this case for example `setColorForElement`

submissions/Roka20012/js-dom/interactive.js
```diff
@@ -92,14 +89,14 @@ function checkStory({ target }) {
     if (target.tagName == "A") {
         for (let i = 0; i < nav.children.length; i++) {
             if (nav.children[i] == target) {
-            	if(!text.children[i].children.length) text.children[i].innerHTML = stories[i + 1];
-                delete stories[i+1];
+                if (!text.children[i].children.length) text.children[i].innerHTML = stories[i + 1];
+                delete stories[i + 1];
                 showText(i);
```
Oodmincheg: do I understand right that here you move your stories[i] to text variable, and when all links in nav was visited your stories are empty and you use only text variable?

Oodmincheg:
> @Oodmincheg Thanks :) I fix it

Please look at the comments below.

submissions/Roka20012/js-dom/interactive.js
Roka20012: Thank you for noticing. I fix it.

submissions/Roka20012/js-dom/interactive.js
Roka20012: Did it)

submissions/Roka20012/js-dom/interactive.js
Roka20012: Taking into account. Thanks

submissions/Roka20012/js-dom/interactive.js
Roka20012: That`s right

Roka20012:
@Oodmincheg Already done)

submissions/Roka20012/js-dom/interactive.js
```diff
@@ -2,52 +2,49 @@
 
 let stories = {
     2: `<p><span><strong>Storie 2.</strong> Lorem ipsum dolor sit amet, consectetur adipisicing elit. Odit similique molestiae eligendi laudantium minus consequatur repellat ipsa, nemo repudiandae perspiciatis, magni aut temporibus nam ipsum expedita aperiam laboriosam quia, sapiente.</span>
```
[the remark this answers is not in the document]
Roka20012: Did it)

submissions/Roka20012/js-dom/interactive.js
Oodmincheg: This is not very good approach. First of all you mutate your input data. Second. You copy your data when it is unnecessarily. Actually you should copy data for avoiding mutating sometimes :)  Third. This is discussable, but the code isn't very clear to understand(to be hones for me very unclear). Maybe we can try to rewrite checkStory and showStory in another way? Just add classes to your navbars links, then add the same names to your data object. and then only change text in your text element. Feel free to ask questions here or in the chat room, if you don't understand something or don't agree with something :)

Oodmincheg:
Please  resolve conversations where you've done.

submissions/Roka20012/js-dom/interactive.js
Oodmincheg: > @Oodmincheg I add comments for clarity.
If it's still not clear I can rewrite it.

Point about clarification was 3d :) There are another 2 reasons why your option isn't nice. 
>I chose this option because the user can click on only 2 menu items and close the site. In my method, every story will only load once and then we just show it to the user and others stories will hide.

If I got you right, you try economy of memory if user click only 1 or 2 items and then close the site. But you already download all your data. so maby this economy unnecessary?

Roka20012: I have transferred all my stories to html div.
Can i do that?or it is`nt nice too?

submissions/Roka20012/js-dom/interactive.js
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * додає колір для вибраного елемента
+ * author: Rostyk Broslavsky
+ * email: broslavsky.rostislaw@gmail.com
+ */
+function setColorForElement(target) {
+    let a = Array.from(nav.children);
+
+    a.forEach((el) => {
```
Oodmincheg: [You can use destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) instead of Array.from. it will be [...nav.children].forEach

submissions/Roka20012/js-dom/interactive.js
Roka20012: @Oodmincheg 
In my opinion, this is not very good because if we have many elements we will need to specify a class or id for each. In my case, I just reassign one id to different elements. I'll think about it again, maybe I'm mistaken thanks.

submissions/Roka20012/js-dom/interactive.js
```diff
@@ -0,0 +1,81 @@
...
+    target.classList.add("check");
+}
+
+/**
+ * при виборі пункта меню показує відповідний текст
+ * author: Rostyk Broslavsky
+ * email: broslavsky.rostislaw@gmail.com
+ */
+function showStory(index) {
+    let textChild = Array.from(text.children);
```
Oodmincheg: In general you shoul try to avoid using of `children`. Because you markup will change sometimes. For examples if you will add  element  inside your `text` element then will be in your array too and all will break. You can select only elements what you need for your manipulation. in this example text.querySelecotrAll('p')

Oodmincheg:
Very good! You fix both cases when you was using children and tagName! Before I even tell you about tagName. We are close to the end ;) Frist of all sync your PR code base with your new code. Second give your `a` variable nice name(menuItem for example). And see one comment below about using computingStyle in your js.

submissions/Roka20012/js-dom/interactive.js
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * закриває меню якщо ми натискаємо на пункт меню 
+ * коли ширина екрана менша 900px
+ * author: Rostyk Broslavsky
+ * email: broslavsky.rostislaw@gmail.com
+ */
+function hideBurgerMenu(target) {
+    let computedStyle = getComputedStyle(burger);
+
+    if (computedStyle.left === "15px") {
```
Oodmincheg: We don't want to use computedStyled number in our JS. Cause if minor changes will ocure in our css file, we will have to fix our code. Maybe there is opportunity move this to the some class? something like moved-to-the-left. And than in js compare classes. Actually I see you have already have the class `open` can we use it?

Roka20012: Thank you.  I did not notice first

submissions/Roka20012/js-dom/interactive.js
```diff
@@ -53,8 +54,8 @@ function showStory(index) {
 function checkStory({ target }) {
 
     if (target.tagName === "A") {//перевіряємо чи елемент на який ми натиснули це пункт меню
```
Oodmincheg: I hurried with tagName :D It's better to use `closest` method [read here](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) with event delegation. in this case would be something like this if(target.closest('#navbar') {..}. That is about more universal code for your markup, to avoid an influance of minor changes in your style on your code.

Roka20012: Thanks i fix it

Oodmincheg:
Great job! Your fixes were really fast. One more time about comments - only english in your next tasks. 

And sually, a developer should understand the logic of your code just by looking on it and that is why properly named functions, variables, used methods and etc. are matters. 

If you need a comment to describe what is going on - 99% that code is badly written and 1% that you really need this because of the difficulty of the task.

Roka20012:
@Oodmincheg  Thank you for the great code review) It`s my first :) I will improve myself

PR #24, opened 2019-02-02
dom-api-task

Zakkarat:
### DOM-practice-task

[Demo](https://zakkarat.github.io/DOM_TASK/index.html) |
[Code base](https://github.com/Zakkarat/DOM_TASK)

Please, review.

### DIFF
--- a/submissions/Zakkarat/js-dom/app.js
+++ b/submissions/Zakkarat/js-dom/app.js
@@ -2,108 +2,37 @@ const hamb = document.getElementById('hamburger');
 const cont = document.getElementById('container');
 const header = document.getElementById('header');
 const text = document.getElementById('main-text')
-const jhin = document.getElementById('jhin');
 const topHead = document.getElementById('head');
-const mf = document.getElementById('mf')
-const kaisa = document.getElementById('kaisa');
-const caitlyn = document.getElementById('caitlyn');
-const jhinm = document.getElementById('jhinm');
-const mfm = document.getElementById('mfm')
-const kaisam = document.getElementById('kaisam');
-const caitlynm = document.getElementById('caitlynm');
 const imgChange = document.getElementById('imgChange');
 const desk = document.getElementById('desk');
+const champ = document.querySelectorAll('a')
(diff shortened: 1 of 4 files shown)


Oodmincheg:
Hello! Very beatufil job! Before I'll start review of your code let's fix one bug. It relates to height of devices where your app is displaying. Can you check display on resulition 823x411(Pixer 2 XL in you developer tools) for example? I don't see picture, and I don't see a scroll in that page... Feel free ask me anything here or in the chat room :)

Zakkarat:
@Oodmincheg Thank you for mentioning that bug.  Will it be okay to fix It by adding scroll, as I did it on my last commit?

Oodmincheg:
> @Oodmincheg Thank you for mentioning that bug. Will it be okay to fix It by adding scroll, as I did it on my last commit?

I think it's up to you in this project.

submissions/Zakkarat/js-dom/stylesheet/main.css
```diff
@@ -0,0 +1,229 @@
...
+  background-color: #540a24;
+  transition: 1s;
+}
+
+.jhin-header {
+  background-color: #94152d !important;
+  transition: 1s;
+}
+.jhin-desk-head {
+  background-color: #821d30 !important;
```
Oodmincheg: This is very bad practice use !important attribute in your css. Can you explain why do you use it in so many cases? Actually you have some pattern in your jhin, cait, kai'sa and MF styles. And you can simplified your css with using classes and toggle them in you js-file. Can you think about it?  If you would stuck with this, write to me here or in the telegram and I'll give you some hints.

Zakkarat: Done some improvements with `!important` part of code.
Sorry, but I did not understand the part with patterns, I've already have some functions in my app.js, which manipulate classes or did you mean something else? Something like having 3 main classes to store data and change their attributes according to function?

Oodmincheg: It's ok. If we look in your css and js code we can see a lot of stuff with very similar attributes and similar behaviour. For example all your 'hero' classes have propery `transition` and all your js function add text to DOM-elements. We can move most of it to one class/function, and the parts that are different we can move to variable/classes and operate them with js. 
So for example. in your css you will have class `.hero` and than classes `.hero-blue, .hero-red`. than in your js you will have only ONE function that will operate with dom-element with class `.hero` something like that if(clicked on blue){element.toggle('.hero-blue')}. you can move your text into the array method and add appropriate text in you function from this array. Try to think in this way. Don't rack your brain to long, if you will stuck more than couple hours write me here or in the telegram.

Zakkarat: Done whole thing, but with a little difference - used `style` attribute in HTML, but now it works with only one function. Is it appropriate to use `style` attribute or should I find other ways?

Oodmincheg: Actually this is not very good use styles in your html. But why you don't want to put them into your CSS. I was talking about one function instead of all function in you js, but not instead your functions AND your CSS. you can keep all styles that you need int your CSS. even move colors there for example create classes like .blue .green etc...

submissions/Zakkarat/js-dom/app.js
```diff
@@ -0,0 +1,46 @@
...
+function champAuto(a) {
+  for (let i = 0; i < champ.length / 2; i++) {
+    if (champ[i] == a || champ[i + 4] == a){
+      menu();
+      header.textContent = names[i];
+      text.textContent = texty[i];
+      head.style.backgroundColor = colorHeader[i];
+      desk.style.backgroundColor = colorDesk[i];
+      cont.style.backgroundColor = colorHeader[i];
+      document.body.style.backgroundColor = colorBody[i];;
```
Roka20012: Hello :) Great job, i noticed inline style in your code, this is a bad practice. You do not need to correct it in this case. Just read about the inline style and try not to use them in the future )[Inline style is bad](http://www.lostsaloon.com/technology/why-is-inline-css-bad-is-it-really-that-bad/)

Zakkarat: Okay. Thank you!

Oodmincheg: HEY, NOOOOOOO :D Fix it please. We have talk already a lot about inline style :) 
Don't use inline style changing. It is a bad approach for several reasons: 

 

1. First of all, a browser will apply such styling for each line separately/ Which means, that every such line of code will be a reason for running of calculations of a page and for drawing it,  so you can receive a performance problem. Read about reflow and repaint.  

2. This is an imperative way, you need to write declarative and describe what your code does, not how. This will make your code shorter and easier to maintain. 

3. Reuse of code. Saying, you will need to rotate some other stuff - you will add a similar line to another part of an application. Which is not right because of DRY. 

4. JS for logic, CSS - for styling.

Zakkarat: Oh, sorry for resolving this too quickly. Now replaced those inline styles with my old(but improved) class system.

Oodmincheg: Very nice!

Oodmincheg:
LGTM.

PR #39, opened 2019-02-21
Friends app done

Zakkarat:
### Friends App

[Demo](https://zakkarat.github.io/Friends-app/) |
[Code base](https://github.com/Zakkarat/Friends-app)

Please, review.

### DIFF
--- a/submissions/Zakkarat/Friends-app/app.js
+++ b/submissions/Zakkarat/Friends-app/app.js
@@ -11,9 +11,10 @@ const descAge = document.getElementById("descAge");
 const male = document.getElementById("male");
 const female = document.getElementById("female");
 const close = document.getElementById("close");
+const searcher = document.getElementById("searchMob");
+const headerPopup = document.getElementsByClassName("headerMenu")[0];
 let staticPeople;
 let people;
-
 fetch("https://randomuser.me/api/?results=30&nat=us")
   .then(function(response) {
     if (response.ok) {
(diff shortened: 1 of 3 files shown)


Oodmincheg:
Hello, Zakkarat. Let's talk about UI. This is not my area of expertise(actually no one :) So this is more discussion than call to action :) 1st I think it's better simple "menu" than "filter". Users don't now what filter means imo. 2nd you modal window with menu don't close when you click outside of it. I think users will be really annoyed about such behavior. 3d It's little confusing that you hide your menu even on big media devices, but it's a pretty confusing that you hide your search bar. Cause I think it's a really hard to understand that search field into the menu(it's always on the page where you will search something) and it's really annoying open-search-close each time when you want to look for something. What do you think about this? Feel free to provide any arguments against of mine :)

Roka20012:
Hello) I see this problem)
bandicam20190223190855372

Zakkarat:
Whole menu idea was changed now. Thanks @Roka20012  for mentioning that bug, now seems to me It's fixed.

Roka20012:
When screen size is small and i click on the menu i see this. This is okay?
image

Zakkarat:
> When screen size is small and i click on the menu i see this. This is okay?
> image

It kinda seems very small, what's size is it?

Roka20012:
When I click on the search input.5,5 inches screen

PR #42, opened 2019-02-26
js_dom

Temu4:
JS DOM API

### task_js_dom

[Demo](https://temu4.github.io/weapons-explorer/) |
[Code base](https://github.com/Temu4/frontend-2019-p2p/tree/dom-api-task/submissions/Temu4/js_dom)

Weapons app with playing around DOM API.
Please, review and help to improve it!


submissions/Temu4/js_dom/script.js
```diff
@@ -0,0 +1,76 @@
...
+  machineGuns,
+  sniperRifles,
+  grenades,
+  ammunition
+];
+
+//Generator for content
+const generateContent = (title, img, description) => {
+  topicTitle.textContent = title;
+  topicImg.style.backgroundImage = img;
```
Roka20012: [Inline styles is bad](http://www.lostsaloon.com/technology/why-is-inline-css-bad-is-it-really-that-bad/)
Don't use inline style changing. It is a bad approach for several reasons:

1. First of all, a browser will apply such styling for each line separately/ Which means, that every such line of code will be a reason for running of calculations of a page and for drawing it,  so you can receive a performance problem. Read about reflow and repaint.
2. This is an imperative way, you need to write declarative and describe what your code does, not how. This will make your code shorter and easier to maintain.
3. Reuse of code. Saying, you will need to rotate some other stuff - you will add a similar line to another part of an application. Which is not right because of DRY.
4. JS for logic, CSS - for styling.

Temu4: Ok, to improve this, I should precreate classes in .css file and then toogle them?

submissions/Temu4/js_dom/script.js
```diff
@@ -0,0 +1,76 @@
...
+  topicDescription.textContent = description;
+};
+
+//Create events
+for (let i = 0; i < menuElements.length; i++) {
+  menuElements[i].addEventListener('click', () => {
+    generateContent(titles[i], imgs[i], descriptions[i]);
+    sideMenu.classList.remove('visible');
+  });
+}
```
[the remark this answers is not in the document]
Temu4: Thank you, for really good advice! I'm fix it in [this commit](https://github.com/kottans/frontend-2019-p2p/pull/42/commits/2172cd6491afcb3508d17461294a44173e296864).

submissions/Temu4/js_dom/script.js
Temu4: Have you some ideas how to handle with inserting images in the content?? 👓

Roka20012: Is up to you,  you can do it with classes if you want :)

Temu4: Fixed by adding new classes. [This](https://github.com/kottans/frontend-2019-p2p/pull/42/commits/ce5a4c4618ff1ead0338b66c8d6c4acbded50df3) commit.

Roka20012:
Nice changes)
Do not forget to mark completed sub-task `resolve coversation`

submissions/Temu4/js_dom/script.js
```diff
@@ -0,0 +1,75 @@
+//Toggle sidemenu
+const sideMenu = document.getElementById('sideMenu');
```
o-msh: Avoid getting element by id. Better way use class.

submissions/Temu4/js_dom/script.js
```diff
@@ -0,0 +1,75 @@
...
+  'Ammunition (informally ammo) is the material fired, scattered, dropped or detonated from any weapon. Ammunition is both expendable weapons (e.g., bombs, missiles, grenades, land mines) and the component parts of other weapons that create the effect on a target (e.g., bullets and warheads). Nearly all mechanical weapons require some form of ammunition to operate.'
+];
+
+//Content elements
+const topicTitle = document.getElementById('topicTitle');
+const topicImg = document.getElementById('topicImg');
+const topicDescription = document.getElementById('topicDescription');
+
+const dataTypes = [
```
o-msh: Try to define you data in common object with pretty structure. Also you repeated types of weapons with `titles` array. DRY.

submissions/Temu4/js_dom/script.js
```diff
@@ -0,0 +1,75 @@
...
+
+//Generator for content
+const generateContent = (title, description) => {
+  topicTitle.textContent = title;
+  topicDescription.textContent = description;
+};
+
+//Create event (in 'Event Delegation' way)
+sideMenu.addEventListener('click', e => {
+  for (let i = 0; i < dataTypes.length; i++) {
```
o-msh: Better way use more declarative approach. `Array` has a great methods.

submissions/Temu4/js_dom/script.js
Temu4: Please, explain?

submissions/Temu4/js_dom/script.js
Temu4: fixed in [this](https://github.com/kottans/frontend-2019-p2p/pull/42/commits/58c5c1a4cfa68b1a1584c140de9a0010d8a48896) commit

submissions/Temu4/js_dom/script.js
Temu4: Fixed. Use forEach() [this commit](https://github.com/kottans/frontend-2019-p2p/pull/42/commits/962f261e6a742e3e69167c5047db7d78f3eceebb)

submissions/Temu4/js_dom/script.js
o-msh: @Temu4  We learn to write code for large and good structured projects. `querySelector` more useful when you want to use more complex selectors in future. Also i see you try to use BEM for html structure. This approach prefer use class name selector. Also remember that IDs are unique identifiers, and for that reason should have more specific names. You can use them only when they are actually necessary and when a class wouldn't be a better way to identify them. Some examples from your code:
```javascript
<nav class="container__side-menu side-menu" id="sideMenu">
``` 
In this case you have perfect element with `side-menu` class. `id` attribute with repeated `side menu` is not necessary. `id` doesn't explain unique element.
```javascript
<button class="toggle-button" id="toggleButton"> 
```
Just use `querySelector` for access to element. `id` doesn't unique also and is not necessary.

Temu4: @o-msh, thanks for you answer. Let I explain, why I've used such class name:
```html
<nav class="container__side-menu side-menu">
```
```css
.side-menu {
  width: 250px;
  height: auto;
  background-color: transparent;
  color: #efefef;
}
```
is responsible for element itself, but
```css
.container__side-menu {
  margin-top: 25px;
  position: absolute;
  -webkit-transform: translate(-300px, 0);
  transform: translate(-300px, 0);
  transition: transform 0.3s ease;
}
```
describes its behavior(position) inside the parent element ("container"). 
I tried separate these and don't mix it, for future performance (if I'll need add the same element at another position in the page).
Also, thank you for mentioning the about ```id```, honestly speaking, I don’t like this approach either, but I didn’t know if it’s right to use ```class``` to select an element (because they can be repeated).

o-msh: @Temu4 Better will be use selection by `class`. So `querySelector` will be preferred. Even if you want to add some repeated elements in future, you must check yourself for right BEM with well structured and nested selectors. In our case we have a simple project with sufficient functional.

Temu4: - [x] issue with id fixed in [this](https://github.com/kottans/frontend-2019-p2p/pull/42/commits/5807b3f2b4609bd55aa7bb6ad18d359c187f6132) commit;
- [x] issue with classes fixed in [this](https://github.com/kottans/frontend-2019-p2p/pull/42/commits/503cae9b959bde4a6a1700542f3d3d61b1d87a45) commit.
Anything else?

PR #74, opened 2019-06-04
Tiny JS World - OOP

aymkin:
### Tiny JS World - OOP

second aproach

[Demo](https://aymkin.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/aymkin/a-tiny-JS-world/tree/tiny-js-world-oop-aymkin)

Please, review.

### DIFF
--- a/submissions/aymkin/OOP-JS (tjsw)/index.js
+++ b/submissions/aymkin/OOP-JS (tjsw)/index.js
@@ -28,6 +28,17 @@ Inhobitant.prototype.getBio = function() {
 	}</strong>. ${this.getLimbs()}. ${this.getFriend()} \nEvery day I say: ${this.makeSound()}.`;
 };
 
+Inhobitant.prototype.getLimbs = function() {
+	let limbs = 'The limbs are not defined on my constructor';
+	if (this.hasOwnProperty('legs')) {
+		limbs = `I have ${this.legs} legs`;
+	}
+	if (this.hasOwnProperty('hands')) {
+		limbs += ` and ${this.hands} hands`;
+	}
+	return limbs;
+};
+
 function Animal(species, name, gender, say, friends) {
 	Inhobitant.call(this, species, name, gender, say, friends);
 	this.legs = 4;
@@ -35,28 +46,33 @@ function Animal(species, name, gender, say, friends) {
 
 Animal.prototype = Object.create(Inhobitant.prototype);
 
-Animal.prototype.getLimbs = function() {
-	return `I have ${this.legs} legs`;
-};
-
-function Human(species, name, gender, say, friends) {
-	Inhobitant.call(this, species, name, gender, say, friends);
+function Human(name, gender, say, friends) {
+	Inhobitant.call(this, name, gender, say, friends);
 	this.legs = 2;
 	this.hands = 2;
+	this.species = 'human';
 }
 
 Human.prototype = Object.create(Inhobitant.prototype);
 
-Human.prototype.getLimbs = function() {
-	return `I have ${this.legs} legs and ${this.hands} hands`;
+function Bird(species, name, gender, say) {
+	Inhobitant.call(this, species, name, gender, say);
+	this.legs = 2;
+	this.wings = 2;
+}
+
+Bird.prototype = Object.create(Inhobitant.prototype);
+
+Bird.prototype.getLimbs = function() {
+	return Inhobitant.prototype.getLimbs.call(this) + ` and ${this.wings} wings`;
 };
 
-const woman = new Human('human', 'Galyna', 'female', 'Dear, do you love me?', [
+const woman = new Human('', 'Galyna', 'female', 'Dear, do you love me?', [
 	'man',
 	'cat',
 ]);
 
-const man = new Human('human', 'Alex', 'male', 'Dear, I love borscht', [
+const man = new Human('', 'Alex', 'male', 'Dear, I love borscht', [
 	'woman',
 	'cat',
 ]);
@@ -68,6 +84,13 @@ const cat = new Animal('cat', 'Benjamin', 'male', 'meow-meow', [
 
 const dog = new Animal('dog', 'Charly', 'male', 'woof-woof');
 
-[woman, man, dog, cat].forEach(inhabitant => {
+const a = new Inhobitant('rhino', 'Asswell', 'malefe', 'bla-bla-bla', [
+	'Barmaley',
+	'Kakadu',
+]);
+
+const blueBird = new Bird('parrot', 'Kakadu', 'female', 'popka-durak');
+
+[woman, man, dog, cat, a, blueBird].forEach(inhabitant => {
 	print(inhabitant.getBio());
 });
--- a/submissions/aymkin/OOP-JS (tjsw)/index.js
+++ b/submissions/aymkin/OOP-JS (tjsw)/index.js
@@ -1,10 +1,10 @@
 /* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
 
-Code repository: https://github.com/aymkin/a-tiny-JS-world/tree/tiny-js-world-oop-aymkin
+Code repository: https://github.com/aymkin/a-tiny-JS-world
 Web app: https://aymkin.github.io/a-tiny-JS-world/
    */
 
-function Inhobitant(species, name, gender, friends, say) {
+function Inhobitant(species, name, gender, say, friends = []) {
 	this.species = species;
 	this.name = name;
 	this.gender = gender;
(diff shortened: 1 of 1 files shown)


submissions/aymkin/OOP-JS (tjsw)/index.js
```diff
@@ -0,0 +1,80 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+
+Code repository: https://github.com/aymkin/a-tiny-JS-world/tree/tiny-js-world-oop-aymkin
+Web app: https://aymkin.github.io/a-tiny-JS-world/
+   */
+
+function Inhobitant(species, name, gender, friends, say) {
```
OleksiyRudenko: There are more chances that an inhabitant has something to say than has got any friends. As per requirements every inhabitant says something and not every inhabitant has friends.

By convention mandatory parameters come first, optional follow.

Under current implementation this means that when we do not want an inhabitant to have friends we still need to specify those, i.e.
`let inhabitant = new Inhabitant('giraffe', 'Griggio', 'male', undefined, 'whatever giraffes say');`

Alternative `let inhabitant = new Inhabitant('giraffe', 'Griggio', 'male', 'whatever giraffes say');` where `friends` param comes at the end would look more consistent.

submissions/aymkin/OOP-JS (tjsw)/index.js
```diff
@@ -0,0 +1,80 @@
...
+	this.friends = friends;
+	this.say = say;
+}
+
+Inhobitant.prototype.makeSound = function() {
+	return this.say;
+};
+
+Inhobitant.prototype.getFriend = function() {
+	return this.hasOwnProperty('friends')
```
OleksiyRudenko: Class `Inhobitant` has own property `friends` even if the value for it is not defined. Under current implementation the `else` condition of the ternary operator will not fire ever.
Please, fix analyzing the `.friends` value.

submissions/aymkin/OOP-JS (tjsw)/index.js
```diff
@@ -0,0 +1,80 @@
...
+		? `I have friends: ${this.friends.join(', ')}.`
+		: `I have no friends`;
+};
+
+Inhobitant.prototype.getHandLegs = function() {
```
OleksiyRudenko: Class `Inhobitant` doesn't implement `.legs` and `.hands`. It is counter-SOLID to implement any logic based on properties class doesn't own.

Imagine, you will have to implement more species like rhinos (and tell a number of horns) or birds (and tell a number of wings) or make the hierarchy of classes deeper. Handling of specific properties should be done at an appropriate level of class hierarchy.

Please, fix.

submissions/aymkin/OOP-JS (tjsw)/index.js
```diff
@@ -0,0 +1,80 @@
...
+	let legs = `I have ${this.legs} legs`;
+	return this.hasOwnProperty('hands')
+		? `${legs} and ${this.hands} hands`
+		: `${legs}`;
+};
+Inhobitant.prototype.getBio = function() {
```
OleksiyRudenko: Based on comment above this method shouldn't use `.getHandLegs` as `.getHandLegs` won't be a part of this class anymore.

So, the question is what to do when we need to report hands or legs from child classes?

The answer is method overloading when a child class re-implements parent method, optionally calling that parent method and supplementing it with additional information. Method overloading is an important OOP concept to master.

submissions/aymkin/OOP-JS (tjsw)/index.js
aymkin: Done:
* parameter `friends` in constructors has been moved to the last 
* check is it friends array empty to reflect the right message

submissions/aymkin/OOP-JS (tjsw)/index.js
aymkin: done

submissions/aymkin/OOP-JS (tjsw)/index.js
aymkin: add the checks if the friend array is empty

submissions/aymkin/OOP-JS (tjsw)/index.js
aymkin: I moved `getHandLeds` to child classes without overloading, should I move it back and use OVERLOADING?

aymkin: should we move on to iteration 2 of code review?

submissions/aymkin/OOP-JS (tjsw)/index.js
```diff
@@ -0,0 +1,73 @@
...
+Inhobitant.prototype.getFriend = function() {
+	return this.hasOwnProperty('friends') && this.friends.length > 0
+		? `I have friends: ${this.friends.join(', ')}.`
+		: `I have no friends`;
+};
+
+Inhobitant.prototype.getBio = function() {
+	return `I am ${this.species}, my name is <strong>${
+		this.name
+	}</strong>. ${this.getLimbs()}. ${this.getFriend()} \nEvery day I say: ${this.makeSound()}.`;
```
OleksiyRudenko: Now, here is the problem:
Class `Inhobitant` doesn't keep any info about limbs.
So the following code
```
let a = new Inhobitant('dog','Aaa','male','Woof');
a.getBio();
```
will produce error `Uncaught TypeError: this.getLimbs is not a function`, which means the code is inconsistent.

The above means that you need to employ overloading, i.e. child classes should implement their own `.getBio` methods. To avoid code duplication child class `.getBio` can call parent's `.getBio`, add something and return a composed result.

The pattern of calling parent's method:
`ParentClass.prototype.myMethod.call(this, arg1, arg2, ..)`

submissions/aymkin/OOP-JS (tjsw)/index.js
```diff
@@ -0,0 +1,73 @@
...
+	this.legs = 4;
+}
+
+Animal.prototype = Object.create(Inhobitant.prototype);
+
+Animal.prototype.getLimbs = function() {
+	return `I have ${this.legs} legs`;
+};
+
+function Human(species, name, gender, say, friends) {
```
OleksiyRudenko: `new Human('human',....)` emm... why? We already know it's a human. Why passing `species` at construction time?

You can pass a pre-defined string to a parent class constructor.

submissions/aymkin/OOP-JS (tjsw)/index.js
aymkin: As far as I could... done), check it well, please

submissions/aymkin/OOP-JS (tjsw)/index.js
aymkin: I got something strange...

aymkin: @OleksiyRudenko did I achieve any of noticed points?

aymkin: I do not resolve any of two comments as far as I am not sure I've done right

submissions/aymkin/OOP-JS (tjsw)/index.js
OleksiyRudenko: Inhobitant doesn't define `.limbs` and it shouldn't. But this also means that `Inhobitant.getBio` cannot refer to an inexistent property `.limbs`.

submissions/aymkin/OOP-JS (tjsw)/index.js
```diff
@@ -0,0 +1,96 @@
...
+		: `I have no friends`;
+};
+
+Inhobitant.prototype.getBio = function() {
+	return `I am ${this.species}, my name is <strong>${
+		this.name
+	}</strong>. ${this.getLimbs()}. ${this.getFriend()} \nEvery day I say: ${this.makeSound()}.`;
+};
+
+Inhobitant.prototype.getLimbs = function() {
```
OleksiyRudenko: Same story as above. Ihobitant doesn't have `.limbs`, so it cannot define any method to retrieve this.

submissions/aymkin/OOP-JS (tjsw)/index.js
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return limbs;
+};
+
+function Animal(species, name, gender, say, friends) {
+	Inhobitant.call(this, species, name, gender, say, friends);
+	this.legs = 4;
+}
+
+Animal.prototype = Object.create(Inhobitant.prototype);
```
OleksiyRudenko: Animal has `legs` and those should have been printed with `.getBio` method.
This means that `Animal` has to define it's own `.getBio`, where it will call parent class `.getBio` and supplement the output of parent's `.getBio` with info re legs.

submissions/aymkin/OOP-JS (tjsw)/index.js
```diff
@@ -0,0 +1,96 @@
...
+
+function Human(name, gender, say, friends) {
```
OleksiyRudenko: Similar to `Animal` with regards to both `.legs` and `.hands`.

submissions/aymkin/OOP-JS (tjsw)/index.js
```diff
@@ -0,0 +1,96 @@
...
+	Inhobitant.call(this, name, gender, say, friends);
+	this.legs = 2;
+	this.hands = 2;
+	this.species = 'human';
+}
+
+Human.prototype = Object.create(Inhobitant.prototype);
+
+function Bird(species, name, gender, say) {
```
OleksiyRudenko: Similar story.

submissions/aymkin/OOP-JS (tjsw)/index.js
OleksiyRudenko: Please, re-read carefully the initial comment in this thread.

OleksiyRudenko:
> @aymkin Nice try, but we didn't actually move much forward from the past iteration. Please, read comments from the past iteration carefully. Pay special attention to parent-child relations and code snippet examples. When implementing ask yourself if new implementation matches the recommendations.
> Please, do not hesitate asking specific questions.

To make your life easier. Imagine you need to implement a new class `Bird` and you cannot change any code related to `Inhobitant`. A bird must report number of legs and wings. Keeping this task in mind refactor `Animal` and `Human` so that requirements above are met.

PR #78, opened 2019-06-13
js-dom complete

Valkirin:
### JS-DOM

[Demo](https://valkirin.github.io/js-dom/) |
[Code base](https://github.com/Valkirin/js-dom)

Please, review.


submissions/Valkirin/js-dom/script.js
```diff
@@ -0,0 +1,103 @@
...
+container.classList.add("container");
+document.body.appendChild(container);
+
+const content = document.createElement("div");
+content.classList.add("content");
+container.appendChild(content);
+
+const contentTitle = document.createElement("h2");
+contentTitle.classList.add("content_title");
+contentTitle.textContent = habbits[0].name;
```
AMashoshyna: I can see a pattern in the way you change text and description for first element and than for clicked element. Can you think of a way to remove this code duplication?

submissions/Valkirin/js-dom/script.js
```diff
@@ -0,0 +1,103 @@
...
+
+const changeMenu = document.querySelector("ul");
+
+changeMenu.addEventListener("click", evt => {
+  const getList = document.querySelectorAll("li");
+  getList.forEach(element => {
+    element.classList.remove("clicked");
+  });
+
+  habbits.forEach((val, key) => {
```
AMashoshyna: The combination of `forEach` and `if` statement can be replaced by array method `find`. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

submissions/Valkirin/js-dom/script.js
```diff
@@ -0,0 +1,103 @@
...
+    name: "Sharpen saw",
+    img: "./img/saw.jpg",
+    description:
+      "Balance and renew your resources, energy, and health to create a sustainable, long-term, effective lifestyle. It primarily emphasizes exercise for physical renewal, good prayer (meditation, yoga, etc.) and good reading for mental renewal. It also mentions service to society for spiritual renewal."
+  }
+];
+
+const container = document.createElement("div");
+container.classList.add("container");
+document.body.appendChild(container);
```
AMashoshyna: DOM manipulations are considered very expensive in terms of performance. For this reason it is better to keep them to minimum. For example, if you need to add multiple elements to the DOM and change their classes or other attributes, it is better to do this in one operation, instead of adding root element and than making changes to it. Please try to rewrite your code, so that you create and adjust DOM elements and append them to the DOM only when they are ready. Common approach is to create container element, append all children and than append container to the DOM. You may find [DocumentFragment API](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) useful.

Valkirin: What an honor for me. Thank you for all your comments. Today I will begin their decision. Unfortunately, due to lack of time, this may take some time.

AMashoshyna: oh wow =) thanks for your kind words

submissions/Valkirin/js-dom/script.js
Valkirin: Code optimized

submissions/Valkirin/js-dom/script.js
Valkirin: Combination replaced

submissions/Valkirin/js-dom/script.js
Valkirin: DocumentFragment used now. So meny time has passed, but I still did it

Valkirin:
@AMashoshyna looking forward to your comments )

PR #101, opened 2019-10-15
Friends-App

madmaxWMFU:
### Friends-App

[Demo](https://madmaxwmfu.github.io/Friends-App/) |
[Code base](https://github.com/madmaxWMFU/Friends-App)

Please, review.


zonzujiro:
Code must be in this PR.

Pls, make new one or change this.

madmaxWMFU:
@zonzujiro, everything is fixed!)

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
+const allFriends = {
+    currentList: [],
+    changeList: [],
+    countFriends: 52,
```
zonzujiro: Count is a verb, so it fits for function name. Use something `numberOfFriends`, as an example

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+    minAge: 0,
+    maxAge: 150
+};
+
+const friendZone = document.querySelector(".friends-zone");
+const filters = document.querySelector(".filters-input");
+const sorts = document.querySelector(".search-sort-options");
+const resetFilters = document.querySelector(".reset-button");
+
+const getDataApi = async () => {
```
zonzujiro: Also problem with naming. From function name I conclude, that you will receive all API as some entity, not some concrete data.

Please, find better naming.

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+    const response = await fetch(`https://randomuser.me/api/?results=${allFriends.countFriends}`);
+    const data = await response.json();
```
zonzujiro: Please, handle network errors - https://www.tjvantoll.com/2015/09/13/fetch-and-errors/ It is a good thing for any application

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+    allFriends.currentList = data.results;
```
zonzujiro: This is a side-effect which should be avoided. For now it is ok, because you are newbie, but in general, you should try to avoid such things.

As as example, this function must have one responsibility and should only return data, not render them or save them elsewhere. All this things must happen outside by it's parent. 

We had materials about purity and single-responsibility, so you should already know this topics. But fill free to ask here anything regarding this topic.

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+                    <span class="user-card__age">${user.dob.age}</span>
+                </div>`;
+    friendZone.insertAdjacentHTML("beforeEnd", temp);
+}
+
+const cleanFriendsZone = () => friendZone.innerHTML = "";
+
+const renderFriendsList = (allFriends) => allFriends.forEach(friend => drawFriendsCards(friend));
+
+const funcFilterList = {
```
zonzujiro: You should avoid of using short words in variables names. Bcs its nt alwys good to read smth lke ths :)

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+    filterByGender(list, {gender}) { return gender !== "all" ? list.filter((person) => person.gender === gender) : allFriends.currentList},
```
zonzujiro: `return` and `{}` are redundant. Replace it with arrow functions

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+    filterByAge(list, {ageTill, ageTo}) { return list.filter((person) => ageTill <= person.dob.age && person.dob.age <= ageTo)},
+    filterByName(list, {name}) { return name != "" ? list.filter((person) => new RegExp(name, 'i').test(person.name.first)) : null}
+}
+
+const getFilterList = () => {
+    allFriends.changeList = [...allFriends.currentList];
+    const searchValue = {
```
zonzujiro: In what case this inputs can be without value? Maybe better will be provide default value for each of them? 🤔

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+        name: document.querySelector(".search-input-name").value || "",
+        gender: document.querySelector("[name='search-input-gender']:checked").value || "all",
+        ageTill: document.querySelector(".search-input-age1").value || allFriends.minAge,
+        ageTo: document.querySelector(".search-input-age2").value || allFriends.maxAge
+    };
+    let funcName = ["filterByName", "filterByGender", "filterByAge"];
```
zonzujiro: You can use `Object.keys` instead

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+    filterByAge(list, {ageTill, ageTo}) { return list.filter((person) => ageTill <= person.dob.age && person.dob.age <= ageTo)},
+    filterByName(list, {name}) { return name != "" ? list.filter((person) => new RegExp(name, 'i').test(person.name.first)) : null}
+}
+
+const getFilterList = () => {
+    allFriends.changeList = [...allFriends.currentList];
+    const searchValue = {
+        name: document.querySelector(".search-input-name").value || "",
+        gender: document.querySelector("[name='search-input-gender']:checked").value || "all",
+        ageTill: document.querySelector(".search-input-age1").value || allFriends.minAge,
```
zonzujiro: `".search-input-age1"` -> `".search-input-age-start"` 🤔 

Numbers in the name of variables or selectors usually from lack of fantasy 😄

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+        ageTo: document.querySelector(".search-input-age2").value || allFriends.maxAge
+    };
+    let funcName = ["filterByName", "filterByGender", "filterByAge"];
+
+    funcName.forEach((func) => {
+        let newArray = funcFilterList[func](allFriends.changeList, searchValue);
```
zonzujiro: Should be `const`

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+    });
+
+    cleanFriendsZone();
+    renderFriendsList(allFriends.changeList);  
+}
+
+const getSortList = ({target}) => {
+    switch(target.value) {
+        case "nameAsc":
+            allFriends.changeList = allFriends.changeList.sort((a, b) => a.name.first < b.name.first ? -1 : 1);
```
zonzujiro: `asc` and `desc` function can be replaced with one function. Try to figure out it

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+        case "ageDesc":
+            allFriends.changeList = allFriends.changeList.sort((a, b) => b.dob.age - a.dob.age);
+            break;
+    }
+    cleanFriendsZone();
+    renderFriendsList(allFriends.changeList);
+};
+
+const resetSearchValues = () => {
+    filters.options[0].selected = true;
```
zonzujiro: Do not rely on such things because in a such way you are dependent from order of children elements. Children can be reordered and your code will be broken. 

Use selector instead.

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+    document.querySelectorAll("[name='search-input-gender']")[2].checked = true;
+    renderFriendsList(allFriends.currentList);
+};
+
+filters.addEventListener("click", getSortList);
+sorts.addEventListener("change", getFilterList);
+resetFilters.addEventListener("click", resetSearchValues);
+
+window.onload = function() {
+    getDataApi()
```
zonzujiro: It is a good place to make realization of pure functions with single responsibilty, etc.

submissions/madmaxWMFU/Friends-App/js/app.js
madmaxWMFU: I agree, sometimes there’s no fantasy at all for the correct name of variables 😞

zonzujiro: So please change name :)

zonzujiro:
Also, please change name of PR to right one

submissions/madmaxWMFU/Friends-App/js/app.js
madmaxWMFU: @zonzujiro, I have to rewrite the function 'sort by name' inside the construct 'switch'?

madmaxWMFU:
@zonzujiro  can you check my code again?

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,117 @@
...
+                    <span class="user-card__age">${user.dob.age}</span>
+                </div>`;
+    friendZone.insertAdjacentHTML("beforeEnd", temp);
+}
+
+const cleanFriendsZone = () => friendZone.innerHTML = "";
+
+const showErrorMessage = () => friendZone.insertAdjacentHTML("beforeEnd", '<h4 class="friends-zone__error">No internet connection. Please try again!)</h4>');
+
+const renderFriendsList = (allFriends) => allFriends.forEach(friend => drawFriendsCards(friend));
```
zonzujiro: It can be just
```suggestion
const renderFriendsList = (allFriends) => allFriends.forEach(drawFriendsCards);
```

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,117 @@
...
+
+const functionFilterList = {
+    filterByGender: (list, {gender}) => gender !== "all" ? list.filter((person) => person.gender === gender) : allFriends.currentList,
+    filterByAge: (list, {ageTill, ageTo}) => list.filter((person) => ageTill <= person.dob.age && person.dob.age <= ageTo),
+    filterByName: (list, {name}) => name != "" ? list.filter((person) => new RegExp(name, 'i').test(person.name.first)) : null
+}
+
+const getGenderValue = () => Array.from(searchGender).filter(item => item.checked === true)[0].value;
```
zonzujiro: `Array#find`

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,117 @@
...
+const cleanFriendsZone = () => friendZone.innerHTML = "";
+
+const showErrorMessage = () => friendZone.insertAdjacentHTML("beforeEnd", '<h4 class="friends-zone__error">No internet connection. Please try again!)</h4>');
+
+const renderFriendsList = (allFriends) => allFriends.forEach(friend => drawFriendsCards(friend));
+
+const functionFilterList = {
+    filterByGender: (list, {gender}) => gender !== "all" ? list.filter((person) => person.gender === gender) : allFriends.currentList,
+    filterByAge: (list, {ageTill, ageTo}) => list.filter((person) => ageTill <= person.dob.age && person.dob.age <= ageTo),
+    filterByName: (list, {name}) => name != "" ? list.filter((person) => new RegExp(name, 'i').test(person.name.first)) : null
```
zonzujiro: Why do you need regexp here? 🤔

zonzujiro: Pls, return array instead `null` for consistency. In such a way your filter function will always return arrays, without edge cases like:
1. If some result - array
2. If no results - `null`
3. If there is full moon or other condition which another developer should check - `undefined` or else :)

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,117 @@
...
+const getFilterList = () => {
+    sorts.value = "none";
+    allFriends.changeList = [...allFriends.currentList];
+    searchValue.name = searchName.value;
+    searchValue.gender = getGenderValue();
+    searchValue.ageTill = searchAgeTill.value;
+    searchValue.ageTo = searchAgeTo.value;
+
+    Object.keys(functionFilterList).forEach((filter) => {
+        const newArray = functionFilterList[filter](allFriends.changeList, searchValue);
```
zonzujiro: `newArray` is a bad name. How you think, why?

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,117 @@
...
+        if(newArray != null) allFriends.changeList = newArray;
+    });
+
+    cleanFriendsZone();
+    renderFriendsList(allFriends.changeList);  
+}
+
+const sortByName = (option) => allFriends.changeList.sort((a, b) => (option === "nameAsc" ? a.name.first < b.name.first : a.name.first > b.name.first) ? -1 : 1);
```
zonzujiro: Nested ternary operators are bad because they very hard to read. Please, rewrite this part.

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,117 @@
...
+
+const sortByAge = (option) => allFriends.changeList.sort((a, b) => option === "ageAsc" ? a.dob.age - b.dob.age : b.dob.age - a.dob.age);
```
zonzujiro: You don't need ternary here.
```
const sortByAge = (a, b) => a.age - b.age;

sortByAge(firstPerson, secondPerson)
sortByAge(secondPerson, firstPerson)
```

Thats all :)

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,117 @@
...
+
+const getSortList = ({target}) => {
```
zonzujiro: Function called `getSomething` but it is don't return anything. Pls, change the name or function implementation.

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,117 @@
...
+    if(target.value != "none")
+        allFriends.changeList = target.value.indexOf("name") != -1 ? sortByName(target.value) : sortByAge(target.value);
+        cleanFriendsZone();
+        renderFriendsList(allFriends.changeList);
+}
+
+const resetSearchValues = () => {
+    sorts.value = "none";
+    document.querySelector(".search-input-name").value = "";
```
zonzujiro: I thought that you can use variables from above for such cases 🤔

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,117 @@
+const allFriends = {
+    currentList: [],
+    changeList: [],
+    numberOfFriends: 52,
+    minAge: 0,
+    maxAge: 150
```
zonzujiro: 😱

submissions/madmaxWMFU/Friends-App/js/app.js
madmaxWMFU: Probably because it is not clear what it contains

submissions/madmaxWMFU/Friends-App/js/app.js
madmaxWMFU: I wanted to show the ability to use regular expressions

zonzujiro: It is a good experience and part when you trying something new - great.
But usually if you can solve problem without regexp - better solve it without regexp.
Please, try to use string methods here because they more readable and easier to understand.

submissions/madmaxWMFU/Friends-App/js/app.js
madmaxWMFU: Did I do something wrong? 😔😔

suchov: I was wondering too.
I think it's just a joke like that form @zonzujiro 
He is amazed by your age policy from 0 t0 150 )))

zonzujiro: @such7 right :) All ok

madmaxWMFU: 😃

submissions/madmaxWMFU/Friends-App/js/app.js
madmaxWMFU: I rewrote the function, but I didn’t refuse the ternary operator due to the choice of sorting

madmaxWMFU:
@zonzujiro, I fixed all errors according to your recommendations

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+const cleanFriendsZone = () => friendZone.innerHTML = "";
+
+const showErrorMessage = () => friendZone.insertAdjacentHTML("beforeEnd", '<h4 class="friends-zone__error">No internet connection. Please try again!)</h4>');
+
+const renderFriendsList = (allFriends) => allFriends.forEach(drawFriendsCards);
+
+const getFilterList = {
+    filterByGender: (list, {gender}) => gender !== "all" ? list.filter((person) => person.gender === gender) : allFriends.currentList,
+    filterByAge: (list, {ageTill, ageTo}) => list.filter((person) => ageTill <= person.dob.age && person.dob.age <= ageTo),
+    filterByName: (list, {name}) => name != "" ? list.filter((person) => person.name.first.toLowerCase().indexOf(name.toLowerCase()) != -1) : []
```
zonzujiro: Pls see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+}
+
+const getGenderValue = () => Array.from(searchGender).find(item => item.checked === true).value;
```
zonzujiro: What if `Array#find` return `undefined`? 🤔 

Also, please, don't use such way of writing because it is hard to read. 
It is better to save result of `Array#find` to some variable and then handle it in appropriate way.

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+    friendZone.insertAdjacentHTML("beforeEnd", temp);
+}
+
+const cleanFriendsZone = () => friendZone.innerHTML = "";
+
+const showErrorMessage = () => friendZone.insertAdjacentHTML("beforeEnd", '<h4 class="friends-zone__error">No internet connection. Please try again!)</h4>');
+
+const renderFriendsList = (allFriends) => allFriends.forEach(drawFriendsCards);
+
+const getFilterList = {
```
zonzujiro: Please use nouns for names of variables which is not a functions.

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+   
+    Object.keys(getFilterList).forEach((filter) => allFriends.changeList = getFilterList[filter](allFriends.changeList, searchValue));
+
+    cleanFriendsZone();
+    renderFriendsList(allFriends.changeList);  
+}
+
+const sortByName = (a, b) => a.name.first < b.name.first ? -1 : 1;
+
+const getSortNameOptionList = (option) => allFriends.changeList.sort((firstPerson, secondPerson) => option === "nameAsc" ? sortByName(firstPerson, secondPerson) : sortByName(secondPerson, firstPerson));
```
zonzujiro: Looks like you have forgot to use prettier.

image

Please, use it. It is a 4k monitor and even on this monitor it is hard to read.

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+
+const sortByAge = (a, b) => a.dob.age - b.dob.age;
+
+const getSortAgeOptionList = (option) => allFriends.changeList.sort((firstPerson, secondPerson) => option === "ageAsc" ? sortByAge(firstPerson, secondPerson) : sortByAge(secondPerson, firstPerson));
+
+const sortList = ({target}) => {
+    if(target.value != "none")
+        allFriends.changeList = target.value.indexOf("name") != -1 ? getSortNameOptionList(target.value) : getSortAgeOptionList(target.value);
```
zonzujiro: `String#includes`.

Please, check all application for `String#indexOf` and replace it everywhere.
`String#indexOf` is more outdated method nowadays.

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+                        <div class="profile-card__name"><address>Place: ${user.location.street.name}, ${user.location.street.number}</address></div>
+                    </div> 
+                    <span class="user-card__age">${user.dob.age}</span>
+                </div>`;
+    friendZone.insertAdjacentHTML("beforeEnd", temp);
+}
+
+const cleanFriendsZone = () => friendZone.innerHTML = "";
+
+const showErrorMessage = () => friendZone.insertAdjacentHTML("beforeEnd", '<h4 class="friends-zone__error">No internet connection. Please try again!)</h4>');
```
zonzujiro: You have confusing  error message.

Your application can fail for any reason and you will show text which is not related to error and more of this - provides wrong information, because user will have internet connection.

In general, error messages must or describe right reason of fail or not to describe it at all, like: "Something went wrong...". But not lie to a user :)

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+
+const getFilterList = {
+    filterByGender: (list, {gender}) => gender !== "all" ? list.filter((person) => person.gender === gender) : allFriends.currentList,
+    filterByAge: (list, {ageTill, ageTo}) => list.filter((person) => ageTill <= person.dob.age && person.dob.age <= ageTo),
+    filterByName: (list, {name}) => name != "" ? list.filter((person) => person.name.first.toLowerCase().indexOf(name.toLowerCase()) != -1) : []
+}
+
+const getGenderValue = () => Array.from(searchGender).find(item => item.checked === true).value;
+
+const filterList = () => {
```
zonzujiro: Name of function can't be noun. Replace it, please.

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+    sorts.value = "none";
+    allFriends.changeList = [...allFriends.currentList];
+    searchValue.name = searchName.value;
+    searchValue.gender = getGenderValue();
+    searchValue.ageTill = searchAgeTill.value;
+    searchValue.ageTo = searchAgeTo.value;
+   
+    Object.keys(getFilterList).forEach((filter) => allFriends.changeList = getFilterList[filter](allFriends.changeList, searchValue));
```
zonzujiro: I believe that if you move this part to different function, like `applyFilters` or something like this, code will become more readable and easier to understand.

In general, you should split your code to such functions and hide details of realization and in such way describe "what" happening instead of "how".

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+        cleanFriendsZone();
+        renderFriendsList(allFriends.changeList);
+}
+
+const resetSearchValues = () => {
+    sorts.value = "none";
+    searchName.value = "";
+    searchAgeTill.value = allFriends.minAge;
+    searchAgeTo.value = allFriends.maxAge;
+    Array.from(searchGender).find(item => item.value === "all").checked = true;
```
zonzujiro: Same with `Array#find` and way of writing.

madmaxWMFU:
I'm sorry, but I don't understand a bit what I must to do that my code was prettier. Maybe something that:
- not to write inline function code;
- use the right name of functions, values
- to break the large functionality code to more smaller
I don`t ignore!) I want that my code or work was prettier.
Please, can you tell me more about my mistake?

submissions/madmaxWMFU/Friends-App/js/app.js
madmaxWMFU: The `undefined` value will not be returned because by default one of the radioboxes is active

madmaxWMFU:
@zonzujiro, I am afraid 🙈
Can you check my amendments again?

submissions/madmaxWMFU/Friends-App/js/app.js
zonzujiro: Resolved, but not fixed.
Please, can you describe why?

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,153 @@
...
+const getFilterList = {
+    filterByGender: (list, {gender}) => {
+        if(gender !== "all") {
+          return list.filter((person) => person.gender === gender); 
+        } else {
+            return allFriends.currentList;
+        }
+    },
+    filterByAge: (list, {ageTill, ageTo}) => {
+        return list.filter((person) => ageTill <= person.dob.age && person.dob.age <= ageTo);
```
zonzujiro:
```suggestion
        return list.filter(({ dob }) => ageTill <= dob.age && dob.age <= ageTo);
```

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,153 @@
...
+    },
+    filterByName: (list, {name}) => {
+        if(name != "") {
```
zonzujiro:
```suggestion
        if(name) {
```

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,153 @@
...
+    searchValues.ageTo = searchAgeTo.value;
+   
+    applyFilters();
+    cleanFriendsZone();
+    renderFriendsList(allFriends.changeList);  
+}
+
+const sortByName = (a, b) => a.name.first < b.name.first ? -1 : 1;
+
+const getSortByNameList = (option) => {
```
zonzujiro: 👍

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,153 @@
...
+        if(option === "ageAsc") {
+            return sortByAge(firstPerson, secondPerson); 
+        } else {
+            return sortByAge(secondPerson, firstPerson);
+        }
+    });
+}
+
+const renderSortFriendList = ({target}) => {
+    if(target.value != "none")
```
zonzujiro: I see that in some cases you use braces and in some cases - not. This is inconsistence. Please, follow one codestyle.
Also, usually in JS we use braces, because it is easier to read code with them. It is not obligatory, but it is common used approach.

submissions/madmaxWMFU/Friends-App/js/app.js
```diff
@@ -0,0 +1,153 @@
...
+        cleanFriendsZone();
+        renderFriendsList(allFriends.changeList);
+}
+
+const resetSearchValues = () => {
+    sorts.value = "none";
+    searchName.value = "";
+    searchAgeTill.value = allFriends.minAge;
+    searchAgeTo.value = allFriends.maxAge;
+    searchGender.find(item => item.value === "all").checked = true;
```
zonzujiro: As I said before, such constructions can be hard to read.
Please, can you describe the reason, why you prefer such approach instead of variable?

submissions/madmaxWMFU/Friends-App/js/app.js
zonzujiro: And what if someday, someone will change your html file without knowing about logic in JS? He will need to analyze all your project to find a code, which needs this default value.

> Also, please, don't use such way of writing because it is hard to read.

And part about "hard to read"? 🤔

submissions/madmaxWMFU/Friends-App/js/app.js
madmaxWMFU: In general, I tried to reduce as much code as possible, and I did not think that its readability suffers because of this.
I thought that if do not use the construction `else`, then I can't write curly brackets

submissions/madmaxWMFU/Friends-App/js/app.js
madmaxWMFU: 😀

submissions/madmaxWMFU/Friends-App/js/app.js
madmaxWMFU: Absolutely agree!

madmaxWMFU:
About the name of the variable `let getFilterList = {}` , for me, it is a set of functions although, in fact, it is an object
Most likely all errors due to not attentiveness, rush, lack of serious projects and mentors

submissions/madmaxWMFU/Friends-App/js/app.js
zonzujiro: I see. Reducing amount of code by removing brackets is not very effective, just because you will not reduce many :)

Also, hope while this code works well with browser, because I haven't checked, it will provide potential bugs for [babel](https://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABACQKYBt1wOpwE7oAmAFAN4DmeqqsY5iAvIgEQAWGWzANIpdVKkKNEAcma4ChZiJ4BnGOlRgow4AEN0s1DwQBZOCC0B5AG6o8XAL4BKRKQBQ9xIhjBiAQj40Yda6SpQIHhIYCCYlgDcjs4xAPSxiAAqRgAiRgBciCkIIiqGqIh4amCEcAC2LkhUJeZOiIoqoRVMumpQrAB0wFj4iMSt7R1FJeXEtgBUiACMAKIA1ADs1h1QcADKUHg-5GNDqAAO6GoQqMSxADod54RzsTDkPMzM1nV1rsTyisqMDEybIKgXjFEABILy0ehMEQiKIxEEgz5KFRMdSaVCwuFNYQABiidQCQSQAB5CDATIgIEdZLIAHJqMqoBgiNCYHD4IgiRCwKCKBikAAGAE0DIg1FRECYYPJVnhEE0AEbmRAAElIcpAFUs_MsiD0BmMZjwfL1-VM5ksAD5ojEibJNghyBa1eDth1ZOgYCdiIhsTwpohlqsAKr7fbmADCai0Y0Qc14VG8dDdHq9U0DcAAMnAAO4RqOnWyWImxO14B1W4EUBMQjpKQiybAwdrEbjPRAAfhYXcytv2xUQdoAnrzSKQIHAsHhMiJznxByJLJbuCxi7I-2ALYhLHVnETUGUrSDndXBFv7CDi_uKzE1R8FEigcCOywOswdzFMsx3Mxt-_i6STAtPEdUQIA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=7.6.4&externalPlugins=).

As an example, code witch looks like this:

```
if(silent === true)
    	greeting = '';
   	silent = false;
    	num = 0;
```

After transpilation will look like this:

```
  if (silent === true) greeting = '';
  silent = false;
  num = 0;
```
Same behaviour with [prettier](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEAzArlMMCW0AEAEnADYkQDqEATiQCYAUwA5tXHLlM-gLz4A6IABalyggDT5W7eHV74A5IKq06ghZIDOOEghjzUAQxKa4k6AFkI6UwHkAbnGriAvgEp8wflG-58OVAYAQmkOHC43YDYYdGoofCh0MhcAbm9fPz8Aeiz8ABVbABFbJHxC6AV9Gzh8akMoOggAW394tganDN19RJa+C0MYIQA6VHIafAYBoeG6huaGDwAqfABGAFEAagB2N2GYCABlGGpw5kXZuAAHEkMwOAYs-n5h57pNrJxmSUFBNwyMgEGNpdLBeDw+Cd0HB-vFMqFONw+AoFGk4X4QXoDMZTGjMgl0H18AAGNEZaKxeIAHjoOHs+DAt00mgAcoYmnAeApiGRKDR6Ap8LgYLoeMAAAYATWs+EMbHw9hw2gO1AJTQARk58AASYBq-AucUufCWax2RzUMWm6oOJwuAB86XR+CpmhO0GY9r1CLOw00JBw9wYJMkq3wewOAFUrlcnABhQymRb4TZSNhhLh+gNB1YRiAAGQgAHd44mHh4XFSsm7qB7Hc6WOnEcMEHRNBQcEMGBI-vgAPwCECD-ClV1Xer4N0AT1FwGAkHI1FKSn40inChcDokg6rmnHUHtBoyfipcCa9f4MG9Tbgchc6RgVbP9fxeuBOj0sPxfgHgleIGPTJSkEIJBHvHxnSrWl7HtMkoGNfAQHEEAICuXBoE0ZBQDlWsiwABTlBBMJQQx7AgHA6CQkB1TqMAAGsOEOccwDOZAoTMYQYCaEgKCETs4D3O44EOIjOzpTsp2QcBmSo8JTGoGA8LqZgmkMZAjBMDiACtNAADwAIVohiYEOdk4HzcI4HUnFtL0w4zl0ABFdAIHgazNOQ8dqHkqT1UMdUZ2gKirlOWAOzoIZkAADmJTza1MCg6iuKSQoEpxHCotgAEd0BwNglMMFS1KQDTTGQ0wmhwNjqGhcqHLgZzXKskqbOQmB-PCyKkAAJjauodDOONmlUqSoGgKzkOqPJ-OI0q4E3IA)

As you probably know, on a job, 99% of time you writing code, which will be transpiled. So it will produce bugs. 
Even when you run formatter like prettier. And this one one is especially bad, because formatter don't change existing logic and if this will happen... well, yikes. Very unpredictable. Imagine if this will happen on CI. You push your code into repo and somewhere there it will run prettier on it. So locally everything is fine, but on CI - it will fail during PR checks. Or event on production.

Anyway, I strongly suggest you not to use such practice outside your pet-projects and encourage you to make code more DRY. This will definitely will reduce amount of it :)

zonzujiro:
> About the name of the variable let getFilterList = {} , for me, it is a set of functions although, in fact, it is an object

I see. Then `listOfFilters` or just `filters` maybe will fit 🤔

submissions/madmaxWMFU/Friends-App/js/app.js
madmaxWMFU: Probably better to do so:
```javascript 
const allGender = document.querySelector('#all[name="search-input-gender"]');
or
const allGender = document.querySelector('#all');
```
Am I right?

zonzujiro: I meant something like this, actually :)

```javascript
const selectAllGenders = () => {
  const allGendersItem = selectedGenders.find(item => item.value === "all");

  if (allGendersItem) {
    allGendersItem.checked = true;
  }
};

selectAllGenders();
```

zonzujiro:
Well, okay. 

We discussed everything and I happy to say that this task is done.
Great job!

madmaxWMFU:
@zonzujiro, Thank you very much for your review.  I got a lot of things for myself.  I will try to implement all the recommendation in future projects.

PR #105, opened 2019-10-28
tiny js world without oop

karelskiy:
### Tiny JS world

[Demo](https://karelskiy.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/karelskiy/a-tiny-JS-world/blob/gh-pages/index.js)

Please, review.

### DIFF
--- a/submissions/karelskiy/tiny-JS-world/index.js
+++ b/submissions/karelskiy/tiny-JS-world/index.js
@@ -30,10 +30,11 @@ const population = [
         legs: 2,
         hands: 2,
         saying: 'Ehi, buongiorno'
-    }
-    
-];
+    }];
 
-population.forEach(el => {
-    print(`${el['species']}; ${el['name']}; ${el['gender']}; ${el['legs']}; ${el['hands']}; ${el['saying']}`);
-})
+population.forEach(el => print(['species', 'name', 'gender', 'legs', 'hands', 'saying']
+  .filter(prop => el[prop] !== undefined) // test if el has the property defined and unlist the prop if not
+  .map(prop => el[prop]) // convert prop names into their values
+  .join('; '))
+);
+ 
--- a/submissions/karelskiy/tiny-JS-world/index.js
+++ b/submissions/karelskiy/tiny-JS-world/index.js
@@ -1,0 +1,39 @@
+const population = [
+    {
+        species: 'dog',
+        name: 'Toby',
+        gender: 'male',
+        legs: 4,
+        hands: 0,
+        saying: 'woof-woof!'
+    },
+    {
+        species: 'cat',
+        name: 'Mrs. Velasquez',
+        gender: 'female',
+        legs: 4,
+        hands: 0,
+        saying: 'meow-meow!'
+    },
+    {
+        species: 'human',
+        name: 'Enrico',
+        gender: 'male',
+        legs: 2,
+        hands: 2,
+        saying: 'Holla'
+    },
+    {
+        species: 'human',
+        name: 'Angel',
+        gender: 'female',
+        legs: 2,
+        hands: 2,
+        saying: 'Ehi, buongiorno'
+    }
+    
+];
+
+population.forEach(el => {
+    print(`${el['species']}; ${el['name']}; ${el['gender']}; ${el['legs']}; ${el['hands']}; ${el['saying']}`);
+})


submissions/karelskiy/tiny-JS-world/index.js
```diff
@@ -0,0 +1,39 @@
...
+        gender: 'female',
+        legs: 2,
+        hands: 2,
+        saying: 'Ehi, buongiorno'
+    }
+    
+];
+
+population.forEach(el => {
+    print(`${el['species']}; ${el['name']}; ${el['gender']}; ${el['legs']}; ${el['hands']}; ${el['saying']}`);
```
OleksiyRudenko: There is a pattern here.
How about mapping an ordered list of properties into values from `el` object and then joining the result with a delimiter?
This would make code [DRYer](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), better readable and scalable (e.g. when we want to add more properties).

submissions/karelskiy/tiny-JS-world/index.js
```diff
@@ -0,0 +1,39 @@
...
+})
```
OleksiyRudenko: Take a note of the red mark added by GitHub.
It is good to have the file ending with a newline for a number of reasons.
You do not have to check this and fix manually. Make your IDE doing this for you.
Find the relevant setting in your IDE and set is as appropriate.
In IntelliJ IDEA tools it is **Preferences > Editor > General > Ensure line feed at file end on Save**. Other IDEs offer similar.

submissions/karelskiy/tiny-JS-world/index.js
karelskiy: Thanks, will be done.

submissions/karelskiy/tiny-JS-world/index.js
karelskiy: Ok, thanks for advice.

submissions/karelskiy/tiny-JS-world/index.js
```diff
@@ -0,0 +1,36 @@
...
+        species: 'human',
+        name: 'Angel',
+        gender: 'female',
+        legs: 2,
+        hands: 2,
+        saying: 'Ehi, buongiorno'
+    }
+];
+
+population.forEach(el => print(Object.values(el).join('; ')));
```
OleksiyRudenko: Good job.
But imagine your peer adds an inhabitant with a different order of props. E.g.
```
{
  name: 'Becky',
  species: 'catwoman',
  gender: 'female',
  hands: 2,
  legs: 2,
  saying: 'Meow!',
}
```
Your overall output will be somewhat confusing.

Moreover in real life there is no guarantee that order of props in object internal presentation would be the same as in the source code. Another use case is composition. That is when you combine two or more objects into one with `Object.assign()`. You may want props be listed in a different order.

So it is good to manage the order explicitly.

Consider the following fragment:
```
const obj = {
  someProp: 5,
  anotherProp: 6,
  yetAnotherProp: 7,
};

['anotherProp', 'yetAnotherProp', 'someProp'].map(prop => obj[prop]); // result: [6, 7, 5]
```

You will find this technique useful in one of the following tasks.

karelskiy: As i understand, i need to display value of properties and also describe their in a few words?

karelskiy: Also i have little problem on GitHub, but it will be resolve soon.

OleksiyRudenko: > As i understand, i need to display value of properties and also describe their in a few words?

Not necessarily. This is the way to guarantee the order of properties for iteration/output.

submissions/karelskiy/tiny-JS-world/index.js
```diff
@@ -0,0 +1,52 @@
...
+    let result, arr = ['', []];
+    for (let i in obj) {
+        if (MAIN_PROP.indexOf(i) >= 0) {
+            result = `Species: <b>${obj['species']};</b> Name: <b>${obj['name']};</b> Gender: <b>${obj['gender']};</b> Legs: <b>${obj['legs']};</b> Hands: <b>${obj['hands']};</b> Favorite phrase: <b>${obj['saying']};</b> `;
+        } else {
+            arr.push(`<i>${i} <b>${obj[i]};</b></i> `);
+        }
+    }
+    return result + arr.join('');
+}
```
OleksiyRudenko: The above is a somewhat overly refactoring.
The array was a proper container for objects. Array guarantees order of elements.
All you needed to do to optimize this fragment
```
population.forEach(el => print(Object.values(el).join('; ')));
```
was
```
population.forEach(el => print(['species', 'name', 'gender', 'legs', 'hands', 'saying']
  .map(prop => el[prop])
  .join('; '))
);
```

If some props can be missing by design and you want to take this into account just add an extra check:

```
population.forEach(el => print(['species', 'name', 'gender', 'legs', 'hands', 'saying']
  .filter(prop => el[prop] !== undefined) // test if el has the property defined
  .map(prop => el[prop]) // convert prop names into their values
  .join('; '))
);
```

karelskiy: Oh, i finally understand, sorry about my tries.

karelskiy:
> @karelskiy well done!
> 
> A nit: don't expect examples to be just copypastable (I mean the code snippet somewhere in the middle of conversation)

Yes, it's clear, i just didn't understand the task well, sorry about this.

OleksiyRudenko:
@karelskiy no worries. I perhaps didn't explain it well. Ask for clarifications with no doubt

karelskiy:
> @karelskiy no worries. I perhaps didn't explain it well. Ask for clarifications with no doubt

Ok, thank you so much

PR #111, opened 2019-11-06
Add oop task frogger game

dashakim:
### OOP task Frogger game

[Demo](https://dashakim.github.io/frontend-frogger-game/) |
[Code base](https://github.com/dashakim/frontend-2019-p2p/tree/oop-task/submissions/dashakim/oop-task)

Please, review.


submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+Enemy.prototype.update = function (dt) {
+
+    this.x += this.speed * dt
+    if (this.x > END_CANVAS) {
+        this.x = BEGIN_CANVAS
+        //random speed for enemies
+        this.speed = 100 + Math.floor(Math.random() * 222)
+    }
+    // return to start position player when collide with enemy
+    if (player.x > (this.x - 60) && player.x < (this.x + 60) && player.y > (this.y - 60) && player.y < (this.y + 60)) {
```
OleksiyRudenko: Using named constants like `END_CANVAS`, `BEGIN_CANVAS` is great for it gives numbers some semantic meaning and code reviewers like myself and possible contributors therefore can clearly understand what stands behind the numbers.

Unnamed constants like `60` here look mysterious. Please fix here and elsewhere across the code.

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,100 @@
+//game score
+var score = 0
+//Enemy class with update(), render() methods
+var Enemy = function (x, y, speed) {
+    this.x = x
+    this.y = y
+    this.speed = speed
+    this.sprite = 'images/car.png';
```
OleksiyRudenko: There are disputes on whether to use semicolons in Javascript or to skip those. But the certain style should be followed across the entire codebase. Please choose your side and fix semicolons.

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+
+    this.x += this.speed * dt
+    if (this.x > END_CANVAS) {
+        this.x = BEGIN_CANVAS
+        //random speed for enemies
+        this.speed = 100 + Math.floor(Math.random() * 222)
+    }
+    // return to start position player when collide with enemy
+    if (player.x > (this.x - 60) && player.x < (this.x + 60) && player.y > (this.y - 60) && player.y < (this.y + 60)) {
+        player.startPosition();
```
OleksiyRudenko: So your class `Enemy` internal behaviour relies on a global variable. If we decide to split the code base into modules (e.g. `enemy.js`, `player.js`, `app.js`) then this becomes a mystery.
Let every enemy know the player at enemy instance initialization and keep the reference to player inside enemy instance.

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+Enemy.prototype.render = function () {
+    ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
+
+var allEnemies = []
+var enemyLocation = [60, 140, 220]
+
+enemyLocation.forEach(function (y) {
+    allEnemies.push(new Enemy(0, y, 200))
+})
```
OleksiyRudenko: This is a use case for [`Array#map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map).

Basically you want to convert an array of y-coordinates into an array of initialized objects. Try to refactor this part.

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+Player.prototype.startPosition = function () {
+    this.x = START_PLAYER_X
+    this.y = START_PLAYER_Y
+}
+
+Player.prototype.update = function () {
+    if (this.y <= 0) {
+        this.startPosition()
+        score++
+        displayStats()
```
OleksiyRudenko: `score` and `displayStats` are 'guests' from the global scope (yet not constants) and also make class too specific. Not good inside the class.
You might (1) use some function like `addScoreBy(increment)` which would call `displayStats` in its turn and also (2) pass this function to the player's instance at initialization (yes, you can pass function as arguments).

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+Player.prototype.render = function () {
+    ctx.drawImage(Resources.get(this.player), this.x, this.y)
+}
```
OleksiyRudenko: Seems like `render` method from `Enemy` and `Player` classes have much in common.
You are not required to do anything with this but you might earn extra skills if you introduce a parent class for the both of them which would implement common properties and behavior.

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,100 @@
...
+})
+
+//start location for player
+var START_PLAYER_X = 200
+var START_PLAYER_Y = 404
+
+//class player
+var Player = function () {
+    this.startPosition()
+    this.player = 'images/char-cat-girl.png'
```
OleksiyRudenko: Similar property in `Enemy` is named `sprite`. Why different here or not `.enemy` there?

submissions/dashakim/oop-task/js/app.js
dashakim: You totally right!

submissions/dashakim/oop-task/js/app.js
dashakim: I'm agree with your point of view, but I'm not sure that I fixed it correctly

submissions/dashakim/oop-task/js/app.js
dashakim: Yes, my fault, there should be sprite

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,106 @@
...
+  top: 0,
+  bottom: 404,
+  halfCell: 50
+};
+
+var Draw = function() {};
+
+Draw.prototype.render = function() {
+  ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
+};
```
OleksiyRudenko: `Draw.render` refers to the properties `Draw` class doesn't have. It is counter Object Oriented Programming principles for parent class to refer the props that only children classes have. Children can refer props they inherit from parent class. Why that? When you have a deep hierarchy of classes maintenance would become hell. [Encapsulation](https://medium.com/@cancerian0684/what-are-four-basic-principles-of-object-oriented-programming-645af8b43727) and [Separation of Concerns](https://en.wikipedia.org/wiki/Separation_of_concerns) principles are to rescue here.

Luckily these props are common for both `Enemy` and `Player` so we can refactor appropriately.
I would also call the class `Character` or something like this so it looks like a real taxonomy.

Any child class can initialize its parent. Here is the pattern to follow:
```
var Figure = function (x, y, figureName) { this.x = x; this.y = y; this.name = figureName; }
Figure.prototype.constructor = Figure;
Figure.prototype.drawCenter = function () { /* ..... */ }

var Circle = function (x, y, radius) {
  Figure.call(this, x, y, 'Circle'); // some arguments are processed by parent constructor
  this.radius = radius;
}
Circle.prototype = Object.create(Figure.prototype);
Circle.prototype.constructor = Circle;
Circle.prototype.draw = function () {
  this.drawCenter(); // this method is inherited
  /* code to draw circle per se */
}
```

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,106 @@
...
+    player.y > this.y - CANVAS.halfCell &&
+    player.y < this.y + CANVAS.halfCell
+  ) {
+    player.startPosition();
+  }
+};
+
+var Player = function() {
+  this.startPosition();
+  this.sprite = "images/char-cat-girl.png";
```
OleksiyRudenko: Looks like `Player` misses props `x`, `y`. Any class prop should be either inherited or defined explicitly.
Of course in Javascript we may create arbitrary props (like `Player.startPosition()` does here) but we write code for other people to read in the first hand. In our case `Player` would inherit these from `Character`.

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+};
+
+var displayStats = function() {
+  gameScore++;
+  document.getElementById("currentStats").innerHTML = "Score: " + gameScore;
+};
+
+var allEnemies = ENEMY.location.map(y => new Enemy(0, y, 200, player));
+var player = new Player();
```
OleksiyRudenko: Hoisting works in JS, but it's counter human nature to see a variable used before it is defined.
This time both lines are next to each other so no big deal. But as the app develops more code may appear in between and things become much less obvious.

Besides when you switch to ES6 with `const` and `let` you will be forced to define things in proper order. So why not to adhere to this principle right off?

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,106 @@
...
+
+var Player = function() {
+  this.startPosition();
+  this.sprite = "images/char-cat-girl.png";
+};
+Player.prototype = Object.create(Draw.prototype);
+Player.prototype.constructor = Player;
+Player.prototype.update = function() {
+  if (this.y <= 0) {
+    displayStats();
```
OleksiyRudenko: An illegal guest from the outer scope. Pass this method at `Player` instantiation and keep it among `Player` class props.

How to pass and store a function:
```
var SomeClass = function (callback) {
 this.callback = callback;
 // .................
 callback();
}

function doSomething() {
  // ...
}

var instance1 = new SomeClass(doSomething()); // wrong; this will pass the result of doSomething call
var instance2 = new SomeClass(doSomething); // correct; this will pass the function per se
```

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,106 @@
...
+    this.startPosition();
+  }
+};
+
+Player.prototype.startPosition = function() {
```
OleksiyRudenko: We want method names to start with a verb.
While technically `start` is both noun and verb from the context it is clear that here it is used as a noun.
`startMoving`, `goToStart`, `resetPosition`, `restart` - all start with a verb.

submissions/dashakim/oop-task/js/app.js
dashakim: I think I did it the same as previous version. Except now I'm understand that method "call" it's like "super" for ES6. Please check if I did it like you suggested. By the way, thanks for  the detailed explanation!

submissions/dashakim/oop-task/js/app.js
dashakim: I changed the scope for "score", but I must confess, I didn't understand if I need to put method or something else  into variable like in this example:
var instance2 = new SomeClass(doSomething); // correct; this will pass the function per se

submissions/dashakim/oop-task/js/app.js
dashakim: Good point! I was so carried away that I didn’t notice .

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,114 @@
...
+  halfCell: 50
+};
+
+var Character = function(x, y) {
+  this.x = x;
+  this.y = y;
+};
+Character.prototype.constructor = Character;
+Character.prototype.render = function() {
+  ctx.drawImage(Resources.get(this.sprite), this.x, this.y);
```
OleksiyRudenko: Any property referred to in a class MUST be the class' own property or inherited from its parent. This class doesn't have any parents, so...

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,114 @@
...
+Enemy.prototype.update = function(dt) {
+  this.x += this.speed * dt;
+  if (this.x > CANVAS.end) {
+    this.x = CANVAS.begin;
+    //random speed for enemies
+    this.speed = ENEMY.speed + Math.floor(Math.random() * ENEMY.varianceSpeed);
+  }
+  // return to start position player when collide with enemy
+  if (
+    player.x > this.x - CANVAS.halfCell &&
```
OleksiyRudenko: `this.player`

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,114 @@
...
+  } else if (keyUp === "up" && this.y > CANVAS.top) {
+    this.y = this.y - PLAYER.stepY;
+  } else if (keyUp === "down" && this.y < CANVAS.bottom) {
+    this.y = this.y + PLAYER.stepY;
+  }
+};
+
+Player.prototype.displayStats = function() {
+  document.getElementById("currentStats").innerHTML = "Score: " + this.score;
+};
```
OleksiyRudenko: Count of Player's lives might be a `Player` class prop. But score is something external to it.
That's why it is more consistent to import an external method to manipulate the score. The score value itself stiull can be stored in a global variable and imported method will have access to it.

dashakim: I think I'm stuck. I nearly understand what's need to change, but I don't understand how. 
I should write global method addScoreBy with global variable "score"(it's acceptable for this stage if I'm understand) than I should add method to properties of Player class and it must works

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,125 @@
...
+    this.player.y < this.y + CANVAS.halfCell
+  ) {
+    this.player.goToStart();
+  }
+};
+
+var Player = function(x, y, addScore) {
+  Character.call(this, x, y);
+  this.addScore = addScore;
+  this.sprite = "images/char-cat-girl.png";
```
OleksiyRudenko: Please induce from the similar case with `Enemy` class.
Please do not expect peer reviewer to point at each line of code that requires fixes similar to those already done.

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,125 @@
...
+  Character.call(this, x, y);
+  this.sprite = "images/car.png";
+  this.speed = speed;
+  this.player = player;
+};
+Enemy.prototype = Object.create(Character.prototype);
+Enemy.prototype.constructor = Enemy;
+Enemy.prototype.toRender = function() {
+  this.render();
+};
```
OleksiyRudenko: Why do you need this method? A child class effectively inherits all methods from its parent class. Besides this method is not called anywhere in the code.
Same for `Player.toRender`

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,125 @@
...
+
+function addScoreBy() {
+  return function() {
+    return gameScore++;
+  };
+}
+var count = addScoreBy();
+
+var displayStats = function() {
+  count();
```
OleksiyRudenko: Why introducing `count`? The question is not rhetoric. Need to understand your way of thinking.

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,125 @@
...
+  return (document.getElementById("currentStats").innerHTML =
+    "Score: " + gameScore);
+};
+
+var player = new Player(PLAYER.startX, PLAYER.startY, displayStats);
+var allEnemies = ENEMY.location.map(y => new Enemy(0, y, 200, player));
```
OleksiyRudenko: Yet another mystery number - `200`.

submissions/dashakim/oop-task/js/app.js
dashakim: I just wanted to call function addScoreBy. And I'm back to  the previous version

submissions/dashakim/oop-task/js/app.js
dashakim: Yes, you're right.

submissions/dashakim/oop-task/js/app.js
dashakim: I'm sorry, I didn't understand what I need to change. Oh, I feel so stupid)

dashakim: You mean to minus score after collision?

submissions/dashakim/oop-task/js/app.js
OleksiyRudenko: This is ok. Why not directly but via `count`?

submissions/dashakim/oop-task/js/app.js
OleksiyRudenko: I mean sprite. You passed it over to a parent class constructor in Enemy class and do not follow the same pattern with Player class

submissions/dashakim/oop-task/js/app.js
dashakim: hmm, if I will call addScoreBy() I will get [function], or you mean I should write:
function addScoreBy() {
      return gameScore++;
    };
addScoreBy()

OleksiyRudenko: Try simplifying this fragment as much as possible (including returning a function from `addScoreBy()`) and see what happens. Add complexity and abstraction only when really in need

OleksiyRudenko: BTW,
```
function doSomething() {
}

var a = new WhatEver(doSomething);
```

Is OK. `doSomething` is already a reference.

Please do not mix ES5 with ES6, stick to using `var` in this task.

dashakim: I use const only for constant variable. I thought it's a good practice. I didn't find any "let"
Ok. I think you have had enough of this PR. But one last hint, what actually I should improve?

submissions/dashakim/oop-task/js/app.js
OleksiyRudenko: You can pass a method/function per se as an argument to use/call it later. Once passed as an argument such function is called a [callback function](https://codeburst.io/javascript-what-the-heck-is-a-callback-aba4da2deced). Callbacks are used in real world tasks very often.

In this case you definitely need to pass a method/function.

submissions/dashakim/oop-task/js/app.js
OleksiyRudenko: Let's do some clean up. And I will take a fresh look onto the code.

OleksiyRudenko:
Please resolve all conversations you believe you have fixed things under. This would help to focus on items that still need elaboration.

submissions/dashakim/oop-task/js/app.js
dashakim: Yes, I'm understand callback functions in general, but I didn't use in real case.  I hope I did it right for now

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+    this.player.y > this.y - CANVAS.halfCell &&
+    this.player.y < this.y + CANVAS.halfCell
+  ) {
+    this.player.goToStart();
+  }
+};
+
+var Player = function(x, y,displayStats) {
+  Character.call(this, x, y);
+  this.displayStats = displayStats
```
OleksiyRudenko: This way we bring an external function received via constructor argument into class instance scope. Good.
Need semicolon though ;)

(Just use [prettier](https://prettier.io/) as a plugin with an IDE of your choice and let it do this dirty job for you)

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+  this.sprite = "images/char-cat-girl.png";
+};
+
+Player.prototype = Object.create(Character.prototype);
+Player.prototype.constructor = Player;
+
+Player.prototype.update = function() {
+  if (this.y <= CANVAS.top) {
+    displayStats(addScore);
```
OleksiyRudenko:
```suggestion
    this.displayStats();
```
`this.displayStats` refers to a function declared globally. The difference is that we do not refer a global function from class scope.

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+  } else if (keyUp === "up" && this.y > CANVAS.top) {
+    this.y = this.y - PLAYER.stepY;
+  } else if (keyUp === "down" && this.y < CANVAS.bottom) {
+    this.y = this.y + PLAYER.stepY;
+  }
+};
+var addScore = function() {
+  return gameScore++
+}
+var displayStats = function(add) {
```
OleksiyRudenko: Both `addScore` and `displayStats` are in the same scope (global). They can call each other directly.

```suggestion
var displayStats = function() {
```

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+  add()
```
OleksiyRudenko:
```suggestion
  addScore();
```

submissions/dashakim/oop-task/js/app.js
```diff
@@ -0,0 +1,118 @@
...
+  } else if (keyUp === "right" && this.x < CANVAS.rightSide) {
+    this.x = this.x + PLAYER.stepX;
+  } else if (keyUp === "up" && this.y > CANVAS.top) {
+    this.y = this.y - PLAYER.stepY;
+  } else if (keyUp === "down" && this.y < CANVAS.bottom) {
+    this.y = this.y + PLAYER.stepY;
+  }
+};
+var addScore = function() {
+  return gameScore++
```
OleksiyRudenko: Returned value is never used and `gameScore` is global variable. Making side effects is not the best practice but OK. So there is no need in `return` op.
```suggestion
  gameScore++;
```

OleksiyRudenko:
@dashakim does the latest version of the game work as expected?

dashakim:
Yes, this is the last working version

PR #112, opened 2019-11-06
Add task tiny world

dashakim:
### a tiny world

[Demo](https://dashakim.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/dashakim/frontend-2019-p2p/tree/js-tiny-world/submissions/dashakim/tiny-world)

Please, review.


submissions/dashakim/tiny-world/index.js
```diff
@@ -0,0 +1,64 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+class Inhabitant {
+    constructor(species, name, gender, legs, hands, saying) {
```
OleksiyRudenko: One of the major points of using classes in OOP is benefiting from class inheritance. This way you may define only what is really required for a class. I.e. animals won't have any hands rather than having zero hands.

Also defining classes like `Human`, `Dog` etc you may skip defining species at initialization.

submissions/dashakim/tiny-world/index.js
```diff
@@ -0,0 +1,64 @@
...
+const cat = new Inhabitant('cat', 'Co-co', 'female', '4', '0', 'Meow')
+const catWoman = new Inhabitant('cat-woman', 'Kitty', 'female', '2', '2')
+man.friends = [woman.name, dog.name].join(', ')
+woman.friends = [cat.name]
+cat.friends = [man.name, woman.name].join(', ')
+dog.friends = [man.name, woman.name, cat.name].join(', ')
+catWoman.saying = cat.saying;
+catWoman.friends = cat.friends
+
+const showIn = function (obj) {
```
OleksiyRudenko: This can be effectively a method of a class and method knows all class internals.
Thanks to inheritance the method will work for any child classes.
Thanks to polymorphism you may redefine method behaviour in child classes.
Thanks to inheritance again you may expand a method in child class (i.e. re-use parent class method and add something else to its result specific to the given child class).

submissions/dashakim/tiny-world/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+const man = new Inhabitant('human', 'Leonid', 'male', '2', '2', 'Hello everyone!')
+const woman = new Inhabitant('human', 'Daria', 'female', '2', '2', 'Good day!')
+const dog = new Inhabitant('dog', 'Archi', 'male', '4', '0', 'Woof-woof')
+const cat = new Inhabitant('cat', 'Co-co', 'female', '4', '0', 'Meow')
+const catWoman = new Inhabitant('cat-woman', 'Kitty', 'female', '2', '2')
+man.friends = [woman.name, dog.name].join(', ')
+woman.friends = [cat.name]
+cat.friends = [man.name, woman.name].join(', ')
+dog.friends = [man.name, woman.name, cat.name].join(', ')
```
OleksiyRudenko: This construction keeps friends' names rather than friends per se.
So if any friend changes her/his name the list won't reflect the changes.
Keeping a list of friends will keep things in sync. Of course you will require a class method to convert a list of friends into a comma-delimited string with their names.

BTW, consider the following snippet:
`[obj1, obj2, obj3].map(obj => obj.name).join(', ');`

OleksiyRudenko:
Emm... Is this your first approach to the Tiny JS World?
If yes then we do all things other way round.

dashakim:
> Emm... Is this your first approach to the Tiny JS World?
> If yes then we do all things other way round.

Yes this is was first approach, actually I forgot to open pull request. Sorry(

submissions/dashakim/tiny-world/index.js
```diff
@@ -0,0 +1,64 @@
...
+        result += `${key}: ${obj[key]}; `
+    }
+    return result
+}
+
+print(showIn(man))
+print(showIn(woman))
+print(showIn(cat))
+print(showIn(dog))
+print(showIn(catWoman))
```
OleksiyRudenko: It is disputable whether to use semicolons in Javascript or not. However there should be the certain approach followed across the code base.
Please choose your side and fix accordingly.

OleksiyRudenko:
> Yes this is was first approach, actually I forgot to open pull request. Sorry(

No worries. Please re-check the comments. Ignore those stroke through for now. They will be useful only when you re-visit the Tiny JS World later in TJSW OOP task.

submissions/dashakim/tiny-world/index.js
dashakim: I decided to delete friendship for current task.  I need to re-write all task where friendship would look like more logic

submissions/dashakim/tiny-world/index.js
dashakim: Thank you for that important comment! I chose my side)

submissions/dashakim/tiny-world/index.js
```diff
@@ -8,40 +8,32 @@
...
     }
-    return result
-}
-
-print(showIn(man))
-print(showIn(woman))
-print(showIn(cat))
-print(showIn(dog))
-print(showIn(catWoman))
+    return result;
```
OleksiyRudenko: The following snippet is equivalent to this function body:
```
return Object.keys(obj).map(key => `${i}: ${obj[i]}`).join('; ');
```
Yet another approach:
```
return Object.entries(obj).map(entry => `${entry[0]}: ${entry[1]}`).join('; ');
```
which can be further refactored to be  even more semantic:
```
return Object.entries(obj).map(([key, value]) => `${key}: ${value}`).join('; ');
```

Consider investigating the snippets above and find out the power of [`Object`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#Methods_of_the_Object_constructor) and [`Array instance`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#Instance_methods) methods.

Those make code concise and semantically clear.

submissions/dashakim/tiny-world/index.js
```diff
@@ -8,40 +8,32 @@
...
+  };
+  
+  inhabitant.forEach(i => print(showInfo(i)));
```
OleksiyRudenko: That's nice! Code became [DRYer](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

PR #122, opened 2019-11-25
Add task oop exercise - a tiny world

dashakim:
_Replace task name and links as appropriate and remove this line._

### OOP exercise second approach

[Demo](https://dashakim.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/dashakim/a-tiny-JS-world/blob/master/index.js)

Please, review.

### DIFF
--- a/submissions/dashakim/a-tiny-world2/index.js
+++ b/submissions/dashakim/a-tiny-world2/index.js
@@ -4,13 +4,19 @@ class Inhabitant {
     this.species = props.species;
     this.gender = props.gender;
     this.saying = props.saying;
-    // this.friends = []
+    this.friends = [];
+  }
+  setFriends(friends) {
+    this.friends = friends;
+  }
+  getFriends() {
+    return this.friends.map(friend => friend.name).join(", ");
   }
   toSay() {
     return this.saying;
   }
   toString() {
-    return `I'm ${this.species}, my name is ${this.name}; gender: ${this.gender}. `;
+    return `I'm ${this.species}; My name is ${this.name}; gender: ${this.gender}; `;
   }
 }
 const SayCatMixin = superclass =>
@@ -32,7 +38,7 @@ class Human extends Inhabitant {
       super.toString() +
       `I have ${this.hands} hands and ${
         this.legs
-      } legs. Saying: ${super.toSay()} `
+      } legs; Saying: ${super.toSay()}; My friends: ${super.getFriends()};`
     );
   }
 }
@@ -55,7 +61,10 @@ class Animal extends Inhabitant {
   }
   toString() {
     return (
-      super.toString() + `I have ${this.paws} paws. Saying: ${super.toSay()} `
+      super.toString() +
+      `I have ${
+        this.paws
+      } paws; Saying: ${super.toSay()}; My friends: ${super.getFriends()};`
     );
   }
 }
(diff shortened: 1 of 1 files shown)


submissions/dashakim/a-tiny-world2/index.js
```diff
@@ -0,0 +1,67 @@
...
+  constructor(name, gender) {
+    this.resident = this.constructor.name;
+    this.name = name;
+    this.gender = gender;
+  }
+  toSring() {
+    return Object.entries(this)
+      .map(([prop, value]) => `${prop}: <b>${value}</b>`)
+      .join("; ");
+  }
```
OleksiyRudenko: This method will be inherited by all child classes down to grandchildren and even further. And it would work correctly.

However there is a chance to train a skill of methods overloading. This task is optional. Please let me know if you're gonna to complete it.

So the task per se:

You current implementation relies on the fact that we need all Object properties be communicated to a user. In real business scenarios there will be many properties that we may not want to expose. Imagine, each inhabitant has `age` property that we do not want to expose.

This means we need to list props to print explicitly. This also means that each class must be responsible for printing the props it has knowledge of. Consequently thus means we need to re-define (overload) method `toString` in each sub-class. Yet we can benefit from calling super-class' method to avoid code duplication. `super.methodName()` is to rescue.

Thus we make code well prepared to support more complex taxonomy of Inhabitants and to rely on the structure of information specific to each particular class. This way we follow a principle of Encapsulation that is one of the key principles of OOP.

Can you improve your code accordingly?

Again, you're not required to do so. Just let me know if you take the challenge.

submissions/dashakim/a-tiny-world2/index.js
```diff
@@ -0,0 +1,67 @@
...
+  constructor(name, saying) {
+    super(name, saying);
+  }
+}
+let inhabitants = [
+  new Man("Jack", "I hope you're having a wonderful day"),
+  new Woman("Sara", "Hey Jack, good to see you!"),
+  new Dog("Archi", "male", "Woof-woof!"),
+  new Cat("Felix", "male"),
+  new CatWoman("Selina")
```
OleksiyRudenko: Two things look a bit weird here.

1. `Dog` requires specifying its saying explicitly while `Cat` does not. This is a somewhat inconsistent API. Imagine we have to extend the taxonomy with other animals. Should our user remember what classes require saying to be specified and what do not? There must be a simple rule. E.g. humans require a saying, others do not. Cat-woman can have a default saying if neither specified.
Also note that `CatWoman` constructor provides for a `saying` argument. What will happen if a user creates CatWoman with an individual saying? Perhaps the output will be something the user didn't expect.

2. The spec says that output log for each inhabitant must contain her/his/its species. We can understand whether a human is a Woman or a Man by her/his gender. This leads us to having the `species` property instead of ambiguous `resident`.

OOP is about well designed, maintainable, and scalable data structure.

Please let me know if you're gonna take any of the two challenges above.

submissions/dashakim/a-tiny-world2/index.js
dashakim: Sounds great, yes I'll take the challenge. Also I don't know how to implement the method to add friends without passing into props directly.

submissions/dashakim/a-tiny-world2/index.js
dashakim: Good point! Thank you! I will change this things.

submissions/dashakim/a-tiny-world2/index.js
OleksiyRudenko: Friends is all different story/challenge.

A couple of hints:

You should store friends rather than their names. This means you can add friends from among inhabitants (objects) that already exist at the moment of their addition. So, it must be a dedicated method to call.

From the other hand when listing friends we need just their names.

OleksiyRudenko:
By the way, saying is something that every inhabitant has.
For the sake of data model consistency disregard the order of props to expose.

submissions/dashakim/a-tiny-world2/index.js
dashakim: Unfortunately, hint doesn't help me. I don't know how to start. Would you so kind to help me? There any possible to write method in existing structure?

OleksiyRudenko: Partial class implementation for brevity:
```
class Alien {
  constructor() {  this.friends = [];  }
  addFriends(friends) {  this.friends = friends;  }
  getFriendsNames() {  return this.friends.map(friend => friend.name).join(", ");  }
  toString() { return "whatever" + this.getFriendsNames(); }
}

const alien = new Alien(), cat = new Cat(), dog = new Dog();
alien.addFriends([cat, dog]);
console.log("Profile: " + alien);
```

submissions/dashakim/a-tiny-world2/index.js
```diff
@@ -4,13 +4,19 @@ class Inhabitant {
     this.species = props.species;
     this.gender = props.gender;
     this.saying = props.saying;
-    // this.friends = []
+    this.friends = [];
+  }
+  setFriends(friends) {
+    this.friends = friends;
```
OleksiyRudenko: Dirty trick to make your function accepting both Array and any number of arguments wrapping them into an Array:
```
function doSmth(...params) {
  params = params.flat();
  return params;
}

doSmth([1,2,3]); // [1,2,3]
doSmth(1,2,3); // [1,2,3]
doSmth(1); // [1]
```

PR #127, opened 2019-12-23
Add task memory game

dashakim:
### Memory Game

[Demo](https://dashakim.github.io/memory_game/) |
[Code base](https://github.com/dashakim/frontend-2019-p2p/tree/memory-game/submissions/dashakim/memory-game)

Please, review.


submissions/dashakim/memory-game/script.js
```diff
@@ -0,0 +1,90 @@
...
+const createBackFaceCard = function(div, el) {
+  const backImage = document.createElement("img");
+  backImage.classList.add("back-face");
+  backImage.setAttribute("src", el);
+  backImage.setAttribute("alt", imagesNames(el));
+  div.appendChild(backImage);
+};
+
+const createCard = function(storage) {
+  storage.map(i => {
```
Roman-Halenko: Some things that raise questions here:
1. Function iterates through an array and may create more than one card. So we can rename our function or take another way: move array enumeration beyond function and call that function as a callback in `.map` method.
2. [As we know](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) - calling `map` method returns new array. Is it really important for us in this case, or we could use another array method without redundant functionality?
3. map callback function taking three arguments: element, index, array (in described order). We can name parameters as we want, but due to some conventions more often use **'e'** for `element` and **'i'** for `index` shortcut.

submissions/dashakim/memory-game/script.js
```diff
@@ -0,0 +1,90 @@
...
+  let isMatch = firstCard.dataset.framework === secondCard.dataset.framework;
+  isMatch ? disableCards() : unflipCards();
+}
+
+function disableCards() {
+  firstCard.removeEventListener("click", flipCard);
+  secondCard.removeEventListener("click", flipCard);
+  resetBoard();
+}
+function unflipCards() {
```
Roman-Halenko: `flipBack` may be more clear name. How you think?

submissions/dashakim/memory-game/script.js
```diff
@@ -0,0 +1,90 @@
...
+    null
+  ];
+}
+(function shuffle() {
+  card.forEach(card => {
+    let random = Math.floor(Math.random() * 16);
+    card.style.order = random;
+  });
+})();
+card.forEach(card => card.addEventListener("click", flipCard));
```
Roman-Halenko: Very important thing about event listeners is event delegation. Adding single listener to each element will significantly increase the number of event listeners on your page and their large number will increase the amount of memory consumed by your page. Which will lead to a decrease in productivity. Without delegating events, you would have to re-bind the event on each new element added to the page.
Try to add single event listener to memory cards parent.
Some stuff, that may help you: [Medium](https://medium.com/@stasonmars/%D0%B4%D0%B5%D0%BB%D0%B5%D0%B3%D0%B8%D1%80%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D0%B5-%D1%81%D0%BE%D0%B1%D1%8B%D1%82%D0%B8%D0%B8%CC%86-%D0%B2-javascript-d91cbdd8916a), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Event/target)

dashakim: Yes, you absolutely right about parent's listeners, I used on task dom, but it was just one event element that I needed to change. Now, I have to select both elements and not clear understand how to do it. 
I tried something like this `section.addEventListener('click', flipCard)`
 but it doesn't work with the function "flipCard"
also something like this but it's also wrong   
```section.onclick = event => { let target = event.target; if (!target) return;  flipCard();}```

File: submissions/dashakim/memory-game/images.js [hunk 4]
```diff
@@ -0,0 +1,21 @@
...
+  "img/facebook.svg",
+  "img/tweeter.svg",
+  "img/pinterest.svg",
+  "img/tweeter.svg",
+  "img/behance.svg",
+  "img/skype.svg",
+  "img/dribbble.svg"
+];
+
+images.map(i=>i.split('.').slice(0,1).join('').substr(4))
```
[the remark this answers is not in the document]
dashakim: Yes, my fault I forgot to remove. Thanks that notice that!

submissions/dashakim/memory-game/script.js
```diff
@@ -0,0 +1,90 @@
+const section = document.querySelector("section");
```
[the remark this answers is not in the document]
dashakim: Good  point! I'll use it this useful advice in future!

submissions/dashakim/memory-game/script.js
dashakim: Callbacks it's my weak side. I'm not sure how rewrite it, so I didn't fix it

submissions/dashakim/memory-game/script.js
dashakim: Ok, you're definitely right, but if I put event on hole section, my `flip` class sets only on section. 
So, I rewrited just how I could)

submissions/dashakim/memory-game/script.js
```diff
@@ -1,90 +1,96 @@
...
 
-const createCard = function(storage) {
-  storage.map(i => {
+const createCards = function(storage) {
+  storage.map(e => {
     const divMemoryCard = document.createElement("div");
     divMemoryCard.classList.add("memory-card");
     section.appendChild(divMemoryCard);
-    divMemoryCard.dataset.framework = imagesNames(i);
+    divMemoryCard.dataset.framework = getImagesNamesForAltAttribute(e);
```
Roman-Halenko: Here, as you can see - function works nut just for **'alt'** attribute.

submissions/dashakim/memory-game/script.js
```diff
@@ -1,90 +1,96 @@
-const section = document.querySelector("section");
+const section = document.querySelector("section.memory-game");
 
-const imagesNames = function(currentImg) {
+const getImagesNamesForAltAttribute = function(currentImg) {
```
Roman-Halenko: At this iteration, it's good enough. No need to rewrite it.

I just wanted to pay your attention: `ForAltAttribute` is the redundant part of the function name because describes the narrow applying (single purpose). Also, your function works with a single image, not an array of images. So `getImageName` would be more clear.
By code writing, variables (and functions, etc.) naming talks about your level a much. Since you just started - it will be very useful to take some time and think more about the successful variable/function name. Take it for a good habit - it will pay off.

submissions/dashakim/memory-game/script.js
```diff
@@ -1,90 +1,96 @@
...
   return currentImg
     .split(".")
     .slice(0, 1)
     .join("")
     .substr(4);
 };
 
-const createFrontFaceCard = function(div) {
+const createFrontFaceCard = function(cardsContainer) {
```
Roman-Halenko: Much better! Like it 👍

submissions/dashakim/memory-game/script.js
```diff
@@ -1,90 +1,96 @@
...
-const card = document.querySelectorAll(".memory-card");
+const cards = document.querySelectorAll(".memory-card");
 let lockBoard = false;
 let hasFlippedCard = false;
 let firstCard, secondCard;
 
-function flipCard() {
+const flipCard = e => {
   if (lockBoard) return;
-  if (this === firstCard) return;
```
Roman-Halenko: Since you rid of this check - some things are broken now. Try to play the game and detect what is going wrong.

submissions/dashakim/memory-game/script.js
```diff
@@ -45,8 +45,8 @@ let firstCard, secondCard;
 const flipCard = e => {
   if (lockBoard) return;
   const target = e.target.parentElement;
+  if (target === firstCard) return;
   target.classList.add("flip");
-
   if (!hasFlippedCard) {
     //first click
     hasFlippedCard = true;
```
Roman-Halenko: Your `flipCard` function is almost done now to become a callback for a single event listener, added to a parent node. A few things, we got to do:
- By clicking the card - the `img` element is an event target now. That is ok, but not perfect since the element we should flip is a `div class="memory-card"`. And getting a `parentElement` of an event target doesn't give as 100% guarantee that we get an exact element. Let's resort to CSS "magic" and make always return as the target an element we exactly want. The magic CSS property is - [pointer-events](https://developer.mozilla.org/ru/docs/Web/CSS/pointer-events). Apply `none` value to all nested elements inside of `memory-card`.
- Rewrite function as event target now is `div class="memory-card"` and no more need to getting parent element.
- Add your event listener to `section` element.
- Since we adding an event listener to a section element - we want to prevent any action if clicked it directly (perhaps, somewhere between cards). Write a simple check to detect if `section` been clicked.

Roman-Halenko: A few words about method instability, when you expect the correct element by relying on nesting deepness. Short [video](https://youtu.be/bnFMnXGgnUc) example with adding padding to element describes one of the possible causes.

dashakim: Wow, that's really CSS magic. Thank's a lot, a learned few new stuff after your review. I know, my implementation isn't quite good enough. But at least a clear understand it, I hope)
Would you so kind to check if I'm improve my code just like you recommended?

submissions/dashakim/memory-game/script.js
```diff
@@ -93,4 +94,4 @@ function resetBoard() {
   });
 })();
 
-cards.forEach(card => card.addEventListener("click", flipCard));
+section.addEventListener("click", flipCard);
```
Roman-Halenko: Well done! Now you have a single event listener with the same game functionality.

submissions/dashakim/memory-game/script.js
```diff
@@ -44,7 +44,8 @@ let firstCard, secondCard;
 
 const flipCard = e => {
   if (lockBoard) return;
-  const target = e.target.parentElement;
+  const target = e.target;
+  if(!section === target) return;
```
Roman-Halenko: That check didn't protect your game from adding `flip` class to section element.

Also, since we need a final tiny change, I would like to meet you with a [destructuring](https://developer.mozilla.org/uk/docs/Web/JavaScript/Reference/Operators/%D0%94%D0%B5%D1%81%D1%82%D1%80%D1%83%D0%BA%D1%82%D1%83%D1%80%D0%B8%D0%B7%D0%B0%D1%86%D1%96%D1%8F), which came with ES6. Writing function parameter this way `const flipCard = {target} => {...}` gives you possibility to use `target` without declaring `const target = e.target`. Hope you enjoy it :)

dashakim: Yes, I know about destructuring and even use in in `resetBoard` function, but still not everywhere, good advice!

Roman-Halenko:
I just found a small bug I didn't see before:
Once two different cards flip back - we can not flip again the first clicked card of that pair. Hope it would be the last thing we should fix :)

dashakim:
> I just found a small bug I didn't see before:
> Once two different cards flip back - we can not flip again the first clicked card of that pair. Hope it would be the last thing we should fix :)

Yes, I also found it, but I don't know how to fix it. I think I should rewrite all game.

submissions/dashakim/memory-game/script.js
```diff
@@ -0,0 +1,90 @@
...
+    secondCard.classList.remove("flip");
+    resetBoard();
+  }, 500);
+}
+function resetBoard() {
+  [hasFlippedCard, lockBoard] = [false, false][(firstCard, secondCard)] = [
+    null,
+    null
+  ];
+}
```
Roman-Halenko: Our bug hiding in this function. After it executes - `firstCard` and `secondCard` still not equal to `null`. Can you explain what this code inside of a function does, step by step?

submissions/dashakim/memory-game/script.js
```diff
@@ -0,0 +1,90 @@
...
+function checkForMatch() {
+  let isMatch = firstCard.dataset.framework === secondCard.dataset.framework;
+  isMatch ? disableCards() : unflipCards();
+}
+
+function disableCards() {
+  firstCard.removeEventListener("click", flipCard);
+  secondCard.removeEventListener("click", flipCard);
+  resetBoard();
+}
```
Roman-Halenko: Also, since now is no longer a single event listener on each card - there is nothing to remove. The function now just calling another function - so we don't need it anymore.

Roman-Halenko: At that time, we removed event listeners from flipped pairs to prevent them from clicking. Now we can do it with adding one more check to `flipCard` function: `if (target.classList.contains('flip')) return` (as one of the possible ways).

submissions/dashakim/memory-game/script.js
dashakim: It's destructuring assignment syntax. it's the same like 
```hasFlippedCard = false,
lockBoard = false,
firstCard and secondCard = null

dashakim: Hm, I'm rewrite like this and seems it's work.

Roman-Halenko: First of all - `[(firstCard, secondCard)] = [null, null];` leads to SyntaxError: invalid destructuring target. It's happened because rounded brackets wraps variables.
The second thing - not the best idea to write it to a single line (readability reasons).

PR #131, opened 2020-01-28
a-tiny-JS-world

NastjonkaK:
### A tiny JS world

[Demo](https://nastjonkak.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/NastjonkaK/a-tiny-JS-world/blob/master/index.js)

Please, review.

### DIFF
--- a/submissions/NastjonkaK/a-tiny-JS-world/index.js
+++ b/submissions/NastjonkaK/a-tiny-JS-world/index.js
@@ -1,0 +1,92 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+// Define your objects here
+
+const dog = {
+  type: "dog",
+  name: "Charlie",
+  gender: "male",
+  legs: 4,
+  hands: 0,
+  hasTail: true,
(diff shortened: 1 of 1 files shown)


OleksiyRudenko:
@NastjonkaK 

It is most likely that you copied all files from your original Tiny-JS-World repo including `.git` directory to the task directory in your fork of frontend-2019-p2p.

That's why here codebase is visualized as a submodule.

image

How to fix: remove `.git` directory, add all files, commit & push to update this PR.

NastjonkaK:
@OleksiyRudenko thanks for the comment. You're right I did just that.

submissions/NastjonkaK/a-tiny-JS-world/index.js
```diff
@@ -0,0 +1,92 @@
...
+  saying: cat.saying
+};
+
+function objToString(obj) {
+  let str = "";
+  for (let p in obj) {
+    if (obj.hasOwnProperty(p)) {
+      str += p + ": " + obj[p] + "; ";
+    }
+  }
```
OleksiyRudenko: Consider employing [Object.entries](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries) method. In combination with `#Array.map` and `#Array.join` it would make this fragment of code shorter and semantically cleaner.

NastjonkaK: Thanks for your comments. I used the methods you wrote above.

PR #132, opened 2020-02-25
JS-DOM-Practice by Evgeniy241984

Evgeniy241984:
### JS-DOM-Practice

[Demo] (https://evgeniy241984.github.io/js-dom/)

[Code base] (https://github.com/Evgeniy241984/Evgeniy241984.github.io/tree/master/js-dom)

Please, review.


submissions/evgeniy241984/js-dom/script.js
```diff
@@ -0,0 +1,129 @@
...
+
+nav.addEventListener("click", filmGenerator);
+
+function filmGenerator({ target }) {
+  let filmNameBtn = target.textContent;
+  listOfFilms.forEach(film => {
+    if (film.header === filmNameBtn) {
+      filmPatter(film, content);
+    }
+  });
```
A-Ostrovnyy: Searching the corresponding object by link content is not a good practice. Take a look at [data-attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes)
also, read about [Array.prototype.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find),
and you can write less code with [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)
Using them or not it's up to you.

A-Ostrovnyy:
Good work.
Looks good to me. Let me know if you will rewrite your task otherwise I will merge your PR.

PR #144, opened 2020-05-14
Master tiny js world

dafen173:
### OOP exercise

[Demo](https://dafen173.github.io/oop-tiny-js-world/) |
[Code base](https://github.com/dafen173/frontend-2019-p2p/blob/master-tiny-js-world/submissions/dafen173/a-tiny-js-world2/index.js)

Please, review.


submissions/dafen173/a-tiny-js-world2/index.js
```diff
@@ -0,0 +1,92 @@
...
+    this.hands = 2;
+    this.legs = 2;
+    this.speach = 'Hello, people!';
+  }
+}
+
+class Animal extends Person {
+  constructor(gender, name, species){
+    super(gender, name, species);
+    this.hands = 0;
```
OleksiyRudenko, line 46: The beauty of OOP is that entities that do not need any property just do not have those. Not zero or null, but no property at all.

submissions/dafen173/a-tiny-js-world2/index.js
```diff
@@ -0,0 +1,92 @@
...
+    this.legs = 4;
```
OleksiyRudenko, line 47: Whatever property is shared must be generalized in a common base class. This is one of OOP principles

submissions/dafen173/a-tiny-js-world2/index.js
```diff
@@ -0,0 +1,92 @@
...
+  } 
+  
+  sayAsAnimal(){
+    if (this.species === 'cat') {
+      this.speach = 'Meow';
+    } else if (this.species === 'cat-woman') {
+      this.speach = 'Meow'
+    } else {
+      this.speach = 'Wohw whow whow'
+    }
```
OleksiyRudenko, line 29: Create classes for every species and let them handle what they do.
A base class should not know anything about child classes implementation.
Imagine, you have dozens of species. Current approach will lead to this method growing enermously becoming heavy with `if/else` constructions. No need - let every species class handle this.

Method overloading would come to rescue, but once you have a class per species you won't need this.

OleksiyRudenko (changes requested):
@dafen173 this is nice transition from literals to classes.
However OOP is not only about syntax, it is rather a set of principles to design scalable and maintainable code by means of building optimal hierarchies.
Please revisit [materials on OOP, SOLID etc from the original project](https://github.com/OleksiyRudenko/a-tiny-JS-world#learn-on-your-own)

PR #148, opened 2020-07-22
Added file with my home task TINY- JS-WORLD

xT0PERx:
### Building a Tiny JS World

[Demo](https://xt0perx.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/xT0PERx/frontend-2019-p2p/blob/A-Tiny-JS-World/submissions/xT0PERx/A-Tiny-JS-World/index.js)

Please, review.

### DIFF
--- a/submissions/xT0PERx/A-Tiny-JS-World/index.js
+++ b/submissions/xT0PERx/A-Tiny-JS-World/index.js
@@ -6,7 +6,7 @@
    */
 
 // ======== OBJECTS DEFINITIONS ========
-const inhabitantsOf = [
+const inhabitants = [
   {
     species: 'human',
     name: 'Mike',
@@ -51,25 +51,10 @@ const inhabitantsOf = [
 
 // ======== OUTPUT ========
 
-let species = 'species';
-let name = 'name';
-let gender = 'gender';
-let legs = ' legs';
-let hands = 'hands';
-let saying = 'saying';
-
-inhabitantsOf.forEach((element) => {
-  print(
-    element.species +
-      '; ' +
-      element.name +
-      '; ' +
-      element.gender +
-      '; ' +
-      element.legs +
-      '; ' +
-      element.hands +
-      '; ' +
-      element.saying
-  );
-});
+function printInhabitants(arr) {
+  arr.forEach((element) => {
+    let newOb = Object.values(element).join(';');
+    print(newOb);
+  });
+}
+printInhabitants(inhabitants);
--- a/submissions/xT0PERx/A-Tiny-JS-World/index.js
+++ b/submissions/xT0PERx/A-Tiny-JS-World/index.js
@@ -1,0 +1,75 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
+const inhabitantsOf = [
+  {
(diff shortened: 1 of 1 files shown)


submissions/xT0PERx/A-Tiny-JS-World/index.js
```diff
@@ -0,0 +1,75 @@
...
+];
+
+// ======== OUTPUT ========
+
+let species = 'species';
+let name = 'name';
+let gender = 'gender';
+let legs = ' legs';
+let hands = 'hands';
+let saying = 'saying';
```
OleksiyRudenko, line 59: These variables aren't used. Perhaps you wanted to refer object properties with a `object[varContainingPropertyName]` notation.

submissions/xT0PERx/A-Tiny-JS-World/index.js
```diff
@@ -0,0 +1,75 @@
...
+      '; ' +
+      element.name +
+      '; ' +
+      element.gender +
+      '; ' +
+      element.legs +
+      '; ' +
+      element.hands +
+      '; ' +
+      element.saying
```
OleksiyRudenko, line 73: Code repetition should be avoided. Imagine you've got 100 properties to join into a string.
You may also want to extract object properties into an array in a particular order.
Consider employing [`Array.map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) to convert a list of property names into values from an object and [`Array.join`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join) to build a resulting string.

xT0PERx: I tried to optimize my code. I think it worked. Thanks for the comments, very helpful for a beginner. I'm waiting for new ones, or PR confirmation.

submissions/xT0PERx/A-Tiny-JS-World/index.js
```diff
@@ -0,0 +1,60 @@
...
+];
+
+// ======== OUTPUT ========
+
+function printInhabitants(arr) {
+  arr.forEach((element) => {
+    let newOb = Object.values(element).join(';');
+    print(newOb);
+  });
+}
```
OleksiyRudenko, line 59: You do not need to fix anything here. Just take a note of the below.
Your function is quite specific as per its name. So, it is good to give a parameter some descriptive name like `inhabitants` (there is no problem in having a variable with the same name in the outer scope). This way your peer developer will understand that here is the variable containing inhabitants to print.

No really need to introduce a variable (`newOb`) as it is being used straight in the next statement. This fragment of code might be as short as
```
function printInhabitants(inhabitants) {
  inhabitants.forEach(inhabitant => print(Object.values(inhabitant).join(';')));
}
```

But if you do (introduce a variable) then give it more contextual name, e.g. `inhabitant`.

Avoid too generic naming like `array` or `arr`, `variable`, `obj` etc unless functions are abstract enough to work with generic types.

Also note that your function relies a lot on order of props in each particular object. Different developer may define the next object's props in different order. Even if one defines an object following the pattern, JS doesn't guarantee that props in an object are stored in the order they are defined in.

Look at the fragment below (incomplete for the sake of brevity):
`inhabitant => ['species', 'name', 'gender', 'legs', 'hands', 'saying'].map(propName => inhabitant[propName]).join('; ')`

This way one can control the order of props and also filter out props that aren't required.

xT0PERx: Большое вам спасибо за такое полное объяснение и за новую полезную информацию.

PR #196, opened 2022-07-20
oop tiny js world

DaniaB24:
_Replace task name and links as appropriate and remove this line._

### oop tiny js world

[Demo](https://daniab24.github.io/a-tiny-JS-world/) |
[Code base](https://github.com/DaniaB24/a-tiny-JS-world/blob/master/index.js)

Please, review.

### DIFF
--- a/submissions/DaniaB24/js-oop/index.js
+++ b/submissions/DaniaB24/js-oop/index.js
@@ -1,0 +1,81 @@
+/* Refer to https://github.com/OleksiyRudenko/a-tiny-JS-world for the task details
+   Complete the below for code reviewers' convenience:
+
+   Code repository: _put repo URL here_
+   Web app: _put project's github pages URL here_
+   */
+
+// ======== OBJECTS DEFINITIONS ========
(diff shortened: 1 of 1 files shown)


OleksiyRudenko (changes requested):
Please check the code against the following requirements, fix as appropriate and confirm compliance.

- [ ] Code is [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), which means that whenever you see
  a pattern in your code those should be eliminated as much as possible. Examples:
  - `print(dog); print(cat); etc ... ` should be refactored employing `Array.forEach` as the least
  - `` `${obj.legs}; ${obj.name}; etc...` `` (yes, strings are also code) must be refactored employing appropriate `Array` methods
- [x] Properties used to form an object presentation string must be explicitly listed/specified
- [x] `Object` methods like `keys`, `values`, `entries` shouldn't be used as these do not guarantee any particular order of keys/values
- [ ] Use statement termination semicolons consistently

